PR #21, opened 2014-12-18
[WIP] Add Capistrano

stevenwilkin:
Ported across from `publify`, not tested


alexwllms:
Is this something that we can complete post adviser launch or should it be shelved?

@benlovell

benlovell:
We can close for now.

PR #56, opened 2015-01-06
Add responsive meta tag to RAD head tag

alexwllms:
@benbarnett @benlovell


benbarnett:
Not 100% sure but you may want to move this above the CSS/JS includes to avoid FOUC

alexwllms:
@benbarnett good idea! :postal_horn:

benlovell:
:+1: after you've squashed.

alexwllms:
squashed!

PR #57, opened 2015-01-06
Restructure the rejection form to fix bug 5232

alexwllms:
@benlovell @benbarnett 

I've used the `<form>` tag to wrap the introductory part of the rejection page so that the validation summary is at the top of the page. Is this valid?

### DIFF
--- a/app/views/principals/rejection_form.html.erb
+++ b/app/views/principals/rejection_form.html.erb
@@ -1,7 +1,7 @@
 <div class="l-constrained">
-  <section class="l-pre-qualification-questions">
-    <%= link_to t('rejection.try_again'), prequalify_principals_path, class: 'l-pre-qualification-questions__try-again t-back' %>
-
+  <%= form_for @message, url: contact_path, as: :contact, html: { class: 'l-pre-qualification-questions' },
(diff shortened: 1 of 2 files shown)


app/views/principals/rejection_form.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="l-constrained">
-  <section class="l-pre-qualification-questions">
-    <%= link_to t('rejection.try_again'), prequalify_principals_path, class: 'l-pre-qualification-questions__try-again t-back' %>
-
+  <%= form_for @message, url: contact_path, as: :contact, html: { class: 'l-pre-qualification-questions' },
```
benbarnett: Whilst we're at it, this should also have a class of `form` to get all the styling

benbarnett:
The HTML seems valid, and you could consider the rejection reasoning as part of the form. So yes I think so. Much more accessible having the summary at the top as well.

Don't know if there were any plans to re-use this form, as this change could make that bit slightly more fiddly? Seems good to me though :+1:

benlovell:
:shipit:

PR #67, opened 2015-01-12
5317 Add two questions to pre-qualification page

alexwllms:
@benlovell @benbarnett


app/views/principals/pre_qualification_form.html.erb
```diff
@@ -1,47 +1,111 @@
...
-        <label class="l-pre-qualification-questions__label">
-          <%= f.radio_button :question_3, 0, class: 'l-pre-qualification-questions__input' %>
-          <%= t('registration.answer_no') %>
-        </label>
-      </fieldset>
+    <%= heading_tag(t('registration.heading'), level: 1) %>
+    <p class="l-pre-qualification-questions__intro-text"><%= t('registration.introduction') %></p>
+
+    <%= form_for @prequalification, url: prequalify_principals_path, html: { class: 'form' } do |f| %>
+      <%= f.form_row :question_1 do %>
```
benbarnett: Do we need `errors_for` in this block?

app/views/principals/pre_qualification_form.html.erb
```diff
@@ -1,47 +1,111 @@
...
+          <div class="form__group-item">
+            <%= f.label :question_1 do %>
+              <%= f.radio_button :question_1, 0, class: 'form__group-input' %>
+              <%= t('registration.answer_no') %>
+            <% end %>
+          </div>
+        </fieldset>
+      <% end %>
+
+      <%= f.form_row :question_2 do %>
```
benbarnett: Ditto here and for all questions

app/views/principals/pre_qualification_form.html.erb
benbarnett: Appreciate they're radio buttons but belt and braces - I'd imagine if none were selected (somehow, or if defaults changed later) then the server would return an error

benlovell: The validation messages are not displayed in this page so I don't think this is necessary.

benbarnett: Cool

benbarnett:
:+1:

benlovell:
We'll pick this up on Monday morning and get the specs in shape as agreed Friday.

benlovell:
:+1:

PR #69, opened 2015-01-14
Provides Adviser import capabilities

benlovell:
* Adds `Lookup::Adviser` and basic validation
* `import:advisers` rake task

I'll squash ddc04ce and 7548daf once I get the thumbs :+1:


app/models/lookup/adviser.rb
```diff
@@ -0,0 +1,11 @@
+module Lookup
+  class Adviser < ActiveRecord::Base
+    validates :reference_number, length: { is: 8 }
+
+    validates_presence_of :name
```
tomas-stefano: Curiosity, :). Why in one validation you use the options format and here you called the method name? Why did not do `validates :name, presence: true'?

benlovell: Good point. I added another commit to tidy this up.

tomas-stefano:
:+1:

lshepstone:
:+1:

PR #76, opened 2015-01-19
Ensures presence of answers

benlovell:


alexwllms:
:+1:

spec/forms/pre_qualification_form_spec.rb
```diff
@@ -3,32 +3,33 @@
...
-      question_4: '1'
-    } end
+    let(:params) do
+      {
+        question_1: '1',
+        question_2: '1',
+        question_3: '1',
+        question_4: '1'
+      }
+    end
```
rnalexander: Why would you collapse the other ones but leave this one expanded?

jongilbraith:
:+1:

spec/forms/pre_qualification_form_spec.rb
benlovell: Because it would have made an unsightly long line.

PR #78, opened 2015-01-19
Remove question three and refactor question names

alexwllms:
@benlovell @lshepstone

### DIFF
--- a/app/forms/pre_qualification_form.rb
+++ b/app/forms/pre_qualification_form.rb
@@ -1,15 +1,15 @@
 class PreQualificationForm
   include ActiveModel::Model
 
-  attr_accessor :firm_active_question, :firm_business_model_question, :firm_status_question, :firm_particular_market_question
+  attr_accessor :active_question, :business_model_question, :status_question, :particular_market_question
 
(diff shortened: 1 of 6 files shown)


app/forms/pre_qualification_form.rb
```diff
@@ -1,15 +1,15 @@
 class PreQualificationForm
   include ActiveModel::Model
 
-  attr_accessor :question_1, :question_2, :question_3, :question_4, :question_5
+  attr_accessor :firm_active_question, :firm_business_model_question, :firm_status_question, :firm_particular_market_question
```
benlovell: I reckon we can drop the `firm_` prefix from all of these. Otherwise the naming is spot on.

config/locales/en.yml
```diff
@@ -28,22 +28,19 @@ en:
     answer_yes: "Yes"
     answer_no: "No"
 
-    question_one:
+    firm_active_question:
```
benlovell: Likewise we can drop the prefix from these keys.

alexwllms:
This.

benlovell:
:+1: 
giphy

lshepstone:
:+1:

PR #90, opened 2015-01-23
Fix javascript bug when returning to page one

alexwllms:
Fixes a bug where after selecting hidden options, progressing to the next page, and the clicking the back button the hidden options would remain selected but hide themselves again.

props to @benbarnett 

@benlovell

### DIFF
--- a/app/assets/javascripts/modules/FieldToggleVisibility.js
+++ b/app/assets/javascripts/modules/FieldToggleVisibility.js
@@ -27,9 +27,7 @@ define(['jquery', 'DoughBaseComponent'], function($, DoughBaseComponent) {
     this.$triggers = this.$el.find('[data-dough-field-trigger]');
     this.$targets = this.$el.find('[data-dough-field-target]');
 
-    if (!this.$triggers.filter('[data-dough-field-trigger-type="show"]').is(':checked')) {
-      this.$targets.addClass('is-hidden');
-    }
+    this.hideRelevantTargets();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


benlovell:
:+1:

app/assets/javascripts/modules/FieldToggleVisibility.js
```diff
@@ -64,7 +62,23 @@ define(['jquery', 'DoughBaseComponent'], function($, DoughBaseComponent) {
...
+
+  FieldToggleVisibilityProto.hideRelevantTargets = function() {
+    var self = this;
+
+    this.$triggers.filter('[data-dough-field-trigger-type="show"]').each(function() {
+      var $trigger = $(this),
+        target = $trigger.attr('data-dough-field-trigger'),
+        $target = self.$targets.filter('[data-dough-field-target="' + target + '"]');
+
+      if (!$trigger.is(':checked')) {
```
dougdroper: what do you think about not(:checked)

benbarnett: In fact it might even be easier changing line 70 to:

```
this.$triggers.filter('[data-dough-field-trigger-type="show"]:not(:checked)').each(function() {
```
and then removing this if statement completely.  What dya reckon?

alexwllms: That looks good. I'll implement that and remove the if statement.

benbarnett:
:ship:  :+1:

PR #93, opened 2015-01-23
Rejig fca number validation

jongilbraith:
Prevent further checks when the number hasn't been supplied at all.

### DIFF
--- a/app/models/principal.rb
+++ b/app/models/principal.rb
@@ -6,11 +6,13 @@ class Principal < ActiveRecord::Base
 
   has_one :firm, primary_key: :fca_number, foreign_key: :fca_number
 
+  validates_presence_of :fca_number
+
   validates :fca_number,
-    presence: true,
     uniqueness: true,
     length: { is: 6 },
-    numericality: { only_integer: true }
+    numericality: { only_integer: true },
+    if: :fca_number?
 
   validates :email_address,
     presence: true,
@@ -31,7 +33,7 @@ class Principal < ActiveRecord::Base
 
   validates_acceptance_of :confirmed_disclaimer, accept: true
 
-  validate :match_fca_number
+  validate :match_fca_number, if: :fca_number?
 
   def to_param
     token.parameterize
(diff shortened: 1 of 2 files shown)


app/models/principal.rb
```diff
@@ -6,11 +6,13 @@ class Principal < ActiveRecord::Base
 
   has_one :firm, primary_key: :fca_number, foreign_key: :fca_number
 
+  validates_presence_of :fca_number
```
benlovell: What was the rationale for breaking this out? As far as I know, you can keep it in with the `validates` and `allow_blank` will work as expected.

app/models/principal.rb
```diff
@@ -66,7 +68,7 @@ def associate_firm
   end
 
   def match_fca_number
-    unless Lookup::Firm.exists?(fca_number: self.fca_number)
+    unless self.fca_number.blank? || Lookup::Firm.exists?(fca_number: self.fca_number)
```
benlovell: On the `validate :match_fca_number` you can add an `if: ->{ fca_number.present? }` and take it out of this extra conditional here. Bit cleaner, WDYT?

app/models/principal.rb
jongilbraith: Ah, didn't think it would but it makes sense. Shall rejig.

app/models/principal.rb
jongilbraith: I thought the same and thought I'd tried it and failed.  Will give it another shot.

jongilbraith:
@benlovell as discussed, found that it was still necessary to have a separate presence validation for the fca number.

Other than that, changed ``allow_blank: true`` to ``if: :fca_number?`` and made the change to the custom validation call.

app/models/principal.rb
```diff
@@ -31,7 +33,7 @@ class Principal < ActiveRecord::Base
 
   validates_acceptance_of :confirmed_disclaimer, accept: true
 
-  validate :match_fca_number
+  validate :match_fca_number, if: ->{ fca_number.present? }
```
benlovell: Doesn't `if: :fca_number?` work the same here too?

jongilbraith: :facepalm:

benlovell:
:+1: looks good to me sir! :facepunch: :beetle:

PR #97, opened 2015-01-23
Customise questionnaire error reporting.

jongilbraith:
" In person advice methods is too short (minimum is 1 character)" (and other checkbox selections) is now "In person advice methods - please select at least one".

"Free initial meeting is not included in the list" > "Also Free initial meeting - please select "yes" or "no"".


benlovell:
:+1: :ship:

PR #99, opened 2015-01-23
Ensures name is merged upon creation of Adviser

benlovell:
Prior to this it was merging in 'Temporary Name' as the adviser's name.


alexwllms:
:+1:

PR #100, opened 2015-01-24
Ajax adviser call

alexwllms:
@benlovell 

Will probably need a refactor from @benbarnett and some JS tests. Will add as a ticket for next sprint.


benlovell:
:+1: I'll add the specs.

PR #104, opened 2015-01-24
Add £ symbol and make percentage fields accessible

alexwllms:
@benlovell

Added the form convention to present the £ symbol within the input field. Added aria-labelledby attributes to make the £ and the % input fields more screen reader friendly.

@amansinghb freestyled the text a "little" bit.


benlovell:
:+1: :ship:

PR #106, opened 2015-01-25
Reworks the `Firm` resource to be non-singular

benlovell:
This is ground work for the subsidiary questionnaire/advisers.


benlovell:
It takes the lotion, it merges its own pull request.

PR #110, opened 2015-01-25
Resolves issue displaying and persisting firm

benlovell:
I've noticed some weaknesses in the specs for persisting firm and subsidiary
questionnaires. Will resolve them in the following PR.


alexwllms:
:+1: 

Oh. Right. It's like that is it.

benlovell:
It's like that, and that's the way it is.

PR #117, opened 2015-01-26
Corrects issue with ownership of `Adviser`

benlovell:
Prior to this, when creating an adviser underneath a subsidiary, the adviser
was always assigned to the top-level firm and not the subsidiary as expected.


lshepstone:
LGTM

tomas-stefano:
:+1:

PR #120, opened 2015-01-26
Add new geographical coverage module

alexwllms:
@benlovell @lshepstone

### DIFF
--- a/app/views/advisers/new.html.erb
+++ b/app/views/advisers/new.html.erb
@@ -39,40 +39,42 @@
               <%= f.errors_for :covers_whole_of_uk %>
               <div class="form__group-item">
                 <label class="form__group-input">
-                  <%= f.radio_button :covers_whole_of_uk, 1, class: 't-covers-whole-of-uk', data: { dough_field_trigger_type: 'hide', dough_field_trigger: '1'  } %>
+                  <%= f.radio_button :covers_whole_of_uk, true, class: 't-covers-whole-of-uk', data: { dough_field_trigger_type: 'hide', dough_field_trigger: '1'  } %>
                   <%= t('registration.answer_yes') %>
                 </label>
               </div>
 
               <div class="form__group-item">
                 <label class="form__group-input">
-                  <%= f.radio_button :covers_whole_of_uk, 0, class: 't-covers-whole-of-uk', data: { dough_field_trigger_type: 'show', dough_field_trigger: '1'  } %>
+                  <%= f.radio_button :covers_whole_of_uk, false, class: 't-covers-whole-of-uk', data: { dough_field_trigger_type: 'show', dough_field_trigger: '1'  } %>
                   <%= t('registration.answer_no') %>
                 </label>
               </div>
             <% end %>
           </fieldset>
 
-          <fieldset class="form__group" data-dough-field-target="1">
-            <legend class="l-questionnaire__legend"><%= t('questionnaire.adviser.geographical_coverage.adviser_services_postcode') %></legend>
-            <%= f.form_row :travel_distance do %>
-              <div>
-                <%= f.errors_for :postcode %>
-                <%= f.label :postcode, t('questionnaire.adviser.geographical_coverage.postcode'), class: 'form__label-heading' %>
-                <%= f.text_field :postcode, class: 't-postcode form__input-container--medium', data: { transform: '' } %>
-              </div>
-            <% end %>
+          <% unless @adviser.covers_whole_of_uk %>
+            <fieldset class="form__group" data-dough-field-target="1">
+              <legend class="l-questionnaire__legend"><%= t('questionnaire.adviser.geographical_coverage.adviser_services_postcode') %></legend>
+              <%= f.form_row :postcode do %>
+                <div>
+                  <%= f.errors_for :postcode %>
+                  <%= f.label :postcode, t('questionnaire.adviser.geographical_coverage.postcode'), class: 'form__label-heading' %>
+                  <%= f.text_field :postcode, class: 't-postcode form__input-container--medium', data: { transform: '' } %>
+                </div>
+              <% end %>
 
-            <%= f.form_row :travel_distance do %>
-              <div>
-                <%= f.errors_for :travel_distance %>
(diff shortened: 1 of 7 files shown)


app/views/advisers/new.html.erb
```diff
@@ -27,94 +27,122 @@
...
-      <fieldset class="form__group">
-        <legend class="l-questionnaire__legend"><%= t('questionnaire.adviser.accreditations.heading') %></legend>
-        <p><%= t('questionnaire.adviser.accreditations.description') %></p>
+            <%= f.form_row :travel_distance do %>
+              <div>
+                <%= f.errors_for :travel_distance %>
+                <%= f.label :travel_distance, t('questionnaire.adviser.geographical_coverage.travel_distance'), class: 'form__label-heading' %>
+                <span>
+                  <%= f.select :travel_distance, TravelDistance.all, { include_blank: true }, { class: 't-travel-distance' } %>
+                  <%= t('questionnaire.adviser.geographical_coverage.unit') %></span>
```
benlovell: The closing span is lonely.

lshepstone:
@benlovell @jongilbraith thumbs, please!

jongilbraith:
LGTM :+1:

alexwllms:
Looks pretty good to me too 
i-have-no-idea-what-im-doing-dog

db/migrate/20150128155644_allow_null_for_adviser_covers_whole_of_uk.rb
```diff
@@ -0,0 +1,5 @@
+class AllowNullForAdviserCoversWholeOfUk < ActiveRecord::Migration
+  def change
+    change_column :advisers, :covers_whole_of_uk, :boolean, null: true, default: nil
```
tomas-stefano: I don't know much of the context. But why the default is nil?

spec/support/adviser_page.rb
```diff
@@ -20,7 +23,12 @@ def matched_adviser?(name)
   end
 
   def covers_whole_of_uk(v)
-    key = I18n.t('questionnaire.adviser.geographical_coverage.covers_whole_of_uk_html')
-    v ? check(key) : uncheck(key)
+    if v
```
tomas-stefano: What is 'v'? I would rename this variable to it's represents.

db/migrate/20150128155644_allow_null_for_adviser_covers_whole_of_uk.rb
lshepstone: `nil` 'cause we need the radio buttons that are tied to it to not be selected at all for new advisers. Any other way to ensure this?

jongilbraith: It's probably not good to use db defaults for an application requirement.  Probably wants to be done in an ``after_initialize`` callback.

spec/support/adviser_page.rb
lshepstone: Or get rid of it all together! :)

db/migrate/20150128155644_allow_null_for_adviser_covers_whole_of_uk.rb
tomas-stefano: About the after_initialize would be an idea, yeah. 
@lshepstone I just asked because I saw the commit message of 357a77613cd8588175d1ab43d24a5f4bcd724536 and thought: 'Ok, but why?'. And yes, sometimes i make that same questions with my commit messages, hehe. Someone in the future could make the same questions about it, ;).

spec/support/adviser_page.rb
tomas-stefano: :)

app/views/advisers/new.html.erb
```diff
@@ -25,93 +25,124 @@
...
+                  <%= t('registration.answer_no') %>
+                </label>
+              </div>
             <% end %>
-          </div>
-        <% end %>
-      </fieldset>
+          </fieldset>
+
+          <% unless @adviser.covers_whole_of_uk %>
```
benlovell: `unless @adviser.covers_whole_of_uk?`

spec/support/adviser_page.rb
```diff
@@ -8,6 +8,9 @@ class AdviserPage < SitePrism::Page
   element :confirmed_disclaimer, '.t-confirmed-disclaimer'
   element :submit, '.t-submit'
 
+  element :does_cover_whole_of_uk, '#adviser_covers_whole_of_uk_true'
```
benlovell: `national` makes more sense.

spec/support/adviser_page.rb
```diff
@@ -8,6 +8,9 @@ class AdviserPage < SitePrism::Page
...
+  element :does_not_cover_whole_of_uk, '#adviser_covers_whole_of_uk_false'
```
benlovell: `local` makes more sense.

config/locales/en.yml
```diff
@@ -121,19 +121,24 @@ en:
...
+        description: This is to show customers if your adviser is willing to travel to provide their services.
+        covers_whole_of_uk: Does your adviser provide UK wide coverage?
+        adviser_services_postcode: Please provide details of the area your adviser offers their services?
+        postcode: Adviser's postcode
+        travel_distance: Distance the adviser is willing to travel
+        unit: Miles
+
+      advisers_qualifications:
+        heading: What are your advisers details?
+        description: You will then be asked to complete some basic information about that Adviser.
```
benlovell: Why is `Adviser` title-cased? It isn't anywhere else.

db/migrate/20150128155644_allow_null_for_adviser_covers_whole_of_uk.rb
benlovell: Relying on the DB here isn't necessary as others pointed out. We should replace this with non-null and default the column. This kind of thing is perfect for `after_initialize` as @jongilbraith pointed out.

jongilbraith: Spoke to @lshepstone irl about this and he clarified.  The column previously had a default value set, so this is reinstating the null / absence of a default.

PR #126, opened 2015-01-27
Adds ordering for reference data models

benlovell:
* Uses values from `ID` as they were seeded in the correct order
* Uses `default_scope` for minimal impact (not usually a fan :shit:)


amansinghb:
:+1:

PR #128, opened 2015-01-27
Shows a message if the adviser already exists.

lshepstone:
Regarding error messages, I think it's a good idea to have them in one place, and not in the JS or view (of course), so I've set this up to return the localised error messages from the lookup endpoint, though there other ways to do this.

@alexwllms @benlovell @jongilbraith

### DIFF
--- a/app/assets/javascripts/modules/AdviserAjaxCall.js
+++ b/app/assets/javascripts/modules/AdviserAjaxCall.js
@@ -10,7 +10,6 @@ define(['jquery'], function($) {
         $error = $('[data-notice="error"]'),
         $inputFieldValue = $(this).val(),
         url = $(this).attr('data-url') + $inputFieldValue + ".json",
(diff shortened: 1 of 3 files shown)


alexwllms:
:+1:

app/assets/javascripts/modules/AdviserAjaxCall.js
```diff
@@ -29,11 +28,15 @@ define(['jquery'], function($) {
...
+        404: function(request) {
+          $error.html("<p>" + request.responseJSON.error + "</p>");
         },
 
-        500: function() {
+        409: function(request) {
+          $error.html("<p>" + request.responseJSON.error + "</p>");
+        },
+
+        500: function(request) {
```
benlovell: I reckon this should be the `default:` case instead.

benlovell:
:ship:

PR #132, opened 2015-01-28
Adds statsd gauges for adviser and firm counts

benlovell:
Adds gauges for running totals of registered advisers and firms.

Introduces `registered` scope on `Firm`, since `email_address` is
required, we can guarantee if this attribute is present, the `Firm` can
be determined `registered`.


amansinghb:
:+1:

PR #135, opened 2015-01-29
#5659 Fixes validation for postcode and travel distance.

lshepstone:


benlovell:
Can we stick a quick spec on here so this behaviour doesn't regress?

lshepstone:
@benlovell yep, taking a look at a spec for this, just had to confirm a requirement, because it seems something was missed here.

alexwllms:
LGTM :+1:

app/models/adviser.rb
```diff
@@ -15,12 +15,12 @@ class Adviser < ActiveRecord::Base
   validates :travel_distance,
     presence: true,
     inclusion: { in: TravelDistance.all },
-    unless: :covers_whole_of_uk?
+    if: -> { covers_whole_of_uk === false }
```
benlovell: I changed these lines back to the regular predicate and the specs are all green. Either the specs aren't covering this case well enough, or we don't need threequals for this comparison after all? 

Also one thing to note - their is a spec https://github.com/moneyadviceservice/rad/blob/5659-hidden-error-messages/spec/models/adviser_spec.rb#L40 that covers those validations and whether they fire given certain values for `covers_whole_of_uk`, we should beef that up to cover the case of `nil`, `false` and `true` now we essentially have the tri-state boolean.

spec/models/adviser_spec.rb
```diff
@@ -29,6 +29,26 @@
     end
 
     describe 'geographical coverage' do
+      describe 'does or does not cover whole of uk' do
+        it 'is required' do
+          expect(build(:adviser, covers_whole_of_uk: nil)).to_not be_valid
+        end
+
+        context 'when not selected' do
```
lshepstone: This is essentially the spec to ensure that the "fix" for #5659 doesn't regress - I'd prefer to not have this.

app/models/adviser.rb
```diff
@@ -12,15 +12,18 @@ class Adviser < ActiveRecord::Base
...
   validates_acceptance_of :confirmed_disclaimer, accept: true
 
+  validates :covers_whole_of_uk,
+    inclusion: { in: [true, false] }
+
   validates :travel_distance,
     presence: true,
     inclusion: { in: TravelDistance.all },
-    unless: :covers_whole_of_uk?
+    unless: -> { covers_whole_of_uk or covers_whole_of_uk.nil? }
```
benlovell: Solid. Just a couple minor things. Can you replace `or` with `||` and I reckon it might be worth moving this logic into a predicate method, then you can use it in both places and revert to `unless: :bleh_something?` 

WDYT?

lshepstone: Yep, makes sense, just not sure what to call the method :)

benlovell:
Perfecto. :+1:

PR #136, opened 2015-01-29
Set field order for the Firm model

jongilbraith:
I'm sure someone will be able to suggest a nicer / prettier way to merge those values.


benlovell:
It's the wrong model. The ticket referred to Adviser.

PR #138, opened 2015-01-29
Set field order for the Firm model.

jongilbraith:
Mistakenly did this while addressing the error ordering on the adviser form, but it would seem the errors do still need sorting in the Firm model.

Currently postcode is numbered ahead of town and county.

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -85,6 +85,28 @@ class Firm < ActiveRecord::Base
     parent.present?
   end
 
+  def field_order
+    [
+      :email_address,
+      :telephone_number,
+      :address_line_one,
+      :address_line_two,
+      :address_town,
+      :address_county,
+      :address_postcode,
+      :in_person_advice_methods,
+      :free_initial_meeting,
+      :initial_meeting_duration,
+      :initial_advice_fee_structures,
+      :ongoing_advice_fee_structures,
+      :allowed_payment_methods,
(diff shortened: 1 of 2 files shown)


app/models/firm.rb
```diff
@@ -85,6 +85,28 @@ def subsidiary?
...
+      :address_postcode,
+      :in_person_advice_methods,
+      :free_initial_meeting,
+      :initial_meeting_duration,
+      :initial_advice_fee_structures,
+      :ongoing_advice_fee_structures,
+      :allowed_payment_methods,
+      :minimum_fixed_fee,
+      :percent_total
+    ] +
```
benlovell: Nice job. Maybe from here though we could try:

```ruby
].tap do |ordered_fields|
  ordered_fields << I18n.t(...).keys
  ordered_fields << :investment_sizes
end
```

That's stylistic nit-picking though, but otherwise it just doesn't look particularly 'pretty' ruby, if that makes any sense at all :smile:

jongilbraith: Yeah, agreed, it's not very pretty.  I'll have a play with some approaches along the lines of what you suggest above.

jongilbraith:
@benlovell what you reckon?

The way you mentioned before would require a ``flatten`` tagged into the end of the ``tap`` block, so this feels neater to me.

benlovell:
:+1: :ship: :shipit:

PR #140, opened 2015-02-01
Resolves issue with cut & pasted copy

benlovell:
Prior to this change a duplicate key was used thus the form appeared as:

screen shot 2015-02-01 at 17 23 20


lshepstone:
:+1:

jongilbraith:
:+1:

PR #143, opened 2015-02-02
Very basic task to export firms and advisers to csv

lshepstone:
Could probably be improved with a dedicated `Exporter` that can be used by other tasks later.

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -1,0 +1,79 @@
+require 'csv'
+
+namespace :export do
+  desc 'Export Firms to CSV=(/path/to/csv)'
+  task firms: :environment do
+    unless csv_path = ENV['CSV']
+      puts 'Usage: rake export:firms CSV=/path/to/csv.ext'
+      abort
+    end
+
+    CSV.open(csv_path, 'wb') do |csv|
+      csv << [
+        'ID',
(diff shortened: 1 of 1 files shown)


asmega:
:+1:

jongilbraith:
:+1:

lib/tasks/export.rake
```diff
@@ -0,0 +1,125 @@
...
+        'Pension Transfer Percent',
+        'Long-Term Care Percent',
+        'Equity Release Percent',
+        'Inheritance Tax and Estate Planning Percent',
+        'Wills and Probate percent',
+        'Other Percent',
+        'Investment Sizes'
+      ]
+
+      Firm.order(id: :asc).each do |firm|
```
benlovell: Maybe include `where.not(email_address: nil)` to only dump those firms that completed the questionnaire? Remembering that partially entered firms are also in the database. If they want the partially entered firms in this dataset then obviously leave as-is.

benlovell:
Nice job. Just a small question/comment.

lib/tasks/export.rake
lshepstone: Seems Kemi and Mark were happy with the export yesterday, and Kemi would like us to keep the partial records in.

PR #144, opened 2015-02-03
Default to `false` for "Covers Whole of UK"

lshepstone:
@benlovell @jongilbraith 

Turns out, Dough's so good, @alexwllms needn't do a thing.

Sad to see the "regression spec" go for when `covers_whole_of_uk` is `nil`, but given the defaults, and then the required validation conditional hacks, best to be out with it.

### DIFF
--- a/db/migrate/20150128155644_default_to_nil_for_adviser_covers_whole_of_uk.rb
+++ b/db/migrate/20150128155644_default_to_nil_for_adviser_covers_whole_of_uk.rb
@@ -1,9 +1,0 @@
-class DefaultToNilForAdviserCoversWholeOfUk < ActiveRecord::Migration
-  def up
-    change_column :advisers, :covers_whole_of_uk, :boolean, default: nil
-  end
-
-  def down
-    change_column :advisers, :covers_whole_of_uk, :boolean, default: false
-  end
-end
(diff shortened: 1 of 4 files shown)


benlovell:
SHIP IT

jongilbraith:
:+1:

db/migrate/20150128155644_default_to_nil_for_adviser_covers_whole_of_uk.rb
```diff
@@ -1,9 +0,0 @@
-class DefaultToNilForAdviserCoversWholeOfUk < ActiveRecord::Migration
```
asmega: Why was the migration removed? Unless it's not been deployed I would keep it present.

benlovell: You just answered your own question.

lshepstone: No affect on the production database @asmega :)

asmega: Thanks for the confirmation!

PR #145, opened 2015-02-04
Change postcode field to validate lowercase

alexwllms:
@benlovell @lshepstone @jongilbraith 

I think this is a better approach than forcing the user to enter uppercase, or relying on javascript to convert the characters to uppercase which introduces a possible point of error.

### DIFF
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -22,7 +22,7 @@ class Adviser < ActiveRecord::Base
 
   validates :postcode,
     presence: true,
-    format: { with: /\A[A-Z\d]{1,4} [A-Z\d]{1,3}\z/ },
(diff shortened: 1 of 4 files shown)


alexwllms:
Give it a thumbing boys.

benlovell:
chuck approves

tomas-stefano:
Is this PR suppose to have some specs? (I'm afraid to ask because Chuck Norris has approved) #infame #joke

app/models/adviser.rb
```diff
@@ -46,6 +48,10 @@ def field_order
 
   private
 
+  def upcase_postcode
```
esbel: In order to achieve the same result, I would personally override the setter method, i.e.:

```ruby
def postcode=(postcode)
  write_attribute(:postcode, postcode.upcase)
end
```

In current implementation, postcode will remain as it was provided until firing the validation. In this specific case it probably won't cause any problems, but there're scenarios where it might cause problems (eg. two-step save, where you actually validate the existence of given address against an external service, before dumping the record into the db).

Of course, my note doesn't stop me from :+1:. Just consider if the code is sufficiently future-proof.

benlovell:
@tomas-stefano never question Chuck. :smile: 

But seriously, this stuff is well covered. All we've done is move the upcasing to the back end, where previously it was occurring on the front end.

app/models/adviser.rb
benlovell: The transformation occurring before validation is exactly the right time in this instance since the upcasing occurs solely for validation, and we know the object can't be persisted (well, at least not in the usual code path) without first being validated.

tomas-stefano:
Ok. I believe it. I was just curious, :). But 'when Chuck Norris approves the PR, the PR is merged by itself'. :smile:. :+1:

PR #146, opened 2015-02-04
Inline the styles for the error pages

alexwllms:
@benlovell @benbarnett 

I've inlined the image, our font and the general styles for each error page as per the Github model.


benbarnett:
The code isn't a looker but we can be pretty confident that our error pages are going to look correct whatever the storm they're being displayed for! :+1:

asmega:
:+1:

benlovell:
:+1: we can investigate some kind of rake task that automates the creation of these pages. I've done this before and it would be a nice little exercise for us, and something we can extract for other projects.

Nice job @alexwllms

benbarnett:
That'd be ace

PR #148, opened 2015-02-05
Remove validation from method of advice question

alexwllms:
@jongilbraith @lshepstone

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -203,7 +203,7 @@ en:
       label_five: Post code *
 
     in_person_advice_methods:
-      heading: Does your firm offer advice (through to transaction) in person? *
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -203,7 +203,7 @@ en:
       label_five: Post code *
 
     in_person_advice_methods:
-      heading: Does your firm offer advice (through to transaction) in person? *
```
lshepstone: There might also be an entry that's not longer needed at the top of this file.

```
  in_person_advice_methods:
    too_short: "- please select at least one"
```

benlovell: Good catch. Will remove.

lshepstone:
:+1:

tomas-stefano:
:+1: :ship:

PR #152, opened 2015-02-07
JS Adviser specs

benlovell:
Added some JS specs to cover the adviser matching AJAX calls from the client. Unfortunately phantomjs didn't want to play ball in travis, and caused all kinds of random and non-deterministic failures, regardless of timeout configuration, and judicious waiting where required.

Super-hacky workaround just for now (I promise to revisit) is to not run that particular step when running in travis. When you run the specs locally they'll do the Right Thing :tm: and at least provide some assurance around that feature.


dougdroper:
:+1:

jongilbraith:
:+1:

Will take a look at this myself next week, as I've battled phantomjs in the past and managed some success.

PR #154, opened 2015-02-07
Switches ActiveJob over to the `sidekiq` adapter

benlovell:
We have been using inline processing until now.

Once this gets merged, we'll need to enable a worker dyno in heroku before
deploying these changes to both QA and production environments.


jongilbraith:
:+1:

alexwllms:
LGTM :+1:

PR #156, opened 2015-02-10
Mounts sidekiq web monitor

benlovell:
Uses the same username and password as the admin control panel.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,5 @@
+require 'sidekiq/web'
+
 Rails.application.routes.draw do
   root 'principals#pre_qualification_form'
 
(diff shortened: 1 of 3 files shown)


asmega:
:+1:

esbel:
:+1:

config/routes.rb
```diff
@@ -40,4 +42,10 @@
     end
     resources :principals, only: [:index, :show]
   end
+
+  Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+    Authentication.authenticate(username, password)
+  end if Rails.env.production?
```
jongilbraith: As discussed, stylistically my preference would be to wrap the whole block in the if statement.

jongilbraith:
Other than that small comment :+1:

PR #160, opened 2015-02-16
Add gemset file for RVM users

darokel:
We normally have this for RVM users in the other repos. Would be good to have it here too.


benlovell:
Bundler takes care of this. I can't think of a circumstance where gem sets are useful, but then I don't use RVM.

darokel:
I understand but this is just to keep with consistency and add support for RVM.

lshepstone:
Meh, +1 for `rbenv` - RVM can die. Jokes. Not really.

benlovell:
Despite my ideologies and if this helps other folks using RVM it won't do us any major harm being around. :wink:

darokel:
Sod off ya rbenv heads :smile:

PR #163, opened 2015-02-25
5853 Geocode Advisers

lshepstone:
Upgrades to the latest version of `mas-rad_core` and adds a feature spec to ensure the adviser is scheduled for geocoding.


benlovell:
:ship:

PR #164, opened 2015-02-28
Bumps core for indexing

benlovell:
The dependencies on the job ordering will need to be broken at some
point soon. This spec modification is a workaround for the timebeing.

I'm thinking we'll need to merge the two jobs, or have the geocode job
fanout to indexing upon success. That may make more sense.


lshepstone:
:+1: (tentative for a passing build)

PR #167, opened 2015-03-11
Bumps core for production geocoder caching

benlovell:


lshepstone:
:+1: Sure.

Gemfile.lock
```diff
@@ -152,12 +152,13 @@ GEM
...
     mail (2.6.3)
       mime-types (>= 1.16, < 3)
-    mas-rad_core (0.0.17)
+    mas-rad_core (0.0.19)
       active_model_serializers
       geocoder
       http
       pg
       rails (~> 4.2.0)
+      redis
```
esbel: redis is locked at 3.2.0, while on consumer it's 3.2.1, is it worth bumping as well, just for the sake of consistency?

esbel:
:+1:, regardless of Redis version

Gemfile.lock
benlovell: Seems as though that's down to `sidekiq`, since it's just one patch level I think I'll leave it for now.

esbel: ok, no worries

PR #181, opened 2015-04-21
Add db:migrate step and make the line copy-n-paste-able

munckymagik:
Trying to make as truly trouble free loading experience as possible.

@benlovell @mconnell

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,9 @@ Be sure to remove or modify the `username` attribute.
 
 ```sh
 $ bundle exec rake db:create \
->  && bundle exec rake db:schema:load \
->  && bundle exec rake db:seed
+  && bundle exec rake db:migrate \
+  && bundle exec rake db:schema:load \
+  && bundle exec rake db:seed
 ```
 
 **NOTE** `db:schema:load` loads into both the test and development databases.


mconnell:
My understanding is the migrate guards against the schema being out of date, which I overheard as a problem getting setup, and the schema:load applies it to both dev+test environments :+1:

README.md
```diff
@@ -40,8 +40,9 @@ Be sure to remove or modify the `username` attribute.
 
 ```sh
 $ bundle exec rake db:create \
->  && bundle exec rake db:schema:load \
->  && bundle exec rake db:seed
+  && bundle exec rake db:migrate \
```
benlovell: You can stack these up like `rake db:create db:migrate db:schema:load db:seed`

munckymagik: Yes you are right of course. However, there are some that cannot be stacked in the same call. It's long enough ago that I can't remember the exact issue, I only remember that `rake db:drop db:create && rake db:migrate` must be done in separate invocations otherwise migrations from engines don't get applied. Have you seen this too? Shall we work out a safe split or shall we leave it? @benlovell

benlovell: Ah good catch. `db:create` and `db:migrate` need to be done in a separate context. I've bumped into that before. It's no issue just to leave it as-is :+1:

PR #186, opened 2015-05-01
Add conditional to prevent an ajax call on an empty field

alexwllms:
@benlovell @benbarnett

### DIFF
--- a/app/assets/javascripts/modules/AdviserAjaxCall.js
+++ b/app/assets/javascripts/modules/AdviserAjaxCall.js
@@ -5,41 +5,44 @@ define(['jquery'], function($) {
   var $inputField = $('[data-input]');
 
   $inputField.keyup(function(){
+    if ( $inputField.val() ) {
 
-    var $success = $('[data-notice="success"]'),
-        $error = $('[data-notice="error"]'),
-        $inputFieldValue = $(this).val(),
-        url = $(this).attr('data-url') + $inputFieldValue + ".json",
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/AdviserAjaxCall.js
```diff
@@ -5,41 +5,44 @@ define(['jquery'], function($) {
...
-
-      error: function(request) {
-        $error.removeClass('is-hidden');
-        $success.addClass('is-hidden');
-      },
-
-      statusCode: {
-        404: function(request) {
-          $error.html("<p>" + request.responseJSON.error + "</p>");
+    if ( $inputField.val() !== '' ) {
```
benlovell: Wouldn't `if ( $inputField.val() )` suffice here?

benlovell: Also it would probably make sense to `trim()` too, just incase a space or multiple spaces are entered.

app/assets/javascripts/modules/AdviserAjaxCall.js
```diff
@@ -5,41 +5,44 @@ define(['jquery'], function($) {
...
+
+      var $success = $('[data-notice="success"]'),
+          $error = $('[data-notice="error"]'),
+          $inputFieldValue = $(this).val(),
```
benbarnett: No need for the `$` here as `.val()` just returns a string

PR #192, opened 2015-05-14
Fix delete advisers validation to ensure we only lookup registered firms

munckymagik:
@neoeno @JiggyPete


JiggyPete:
:+1:

neoeno:
Short enough to be a guard clause now I think :trollface:

munckymagik:
Actually it's 135 chars and our limit is 120 so "get tae!" ...

PR #193, opened 2015-05-15
Test RSpec integration: Sign in, reset password.

neoeno:
@munckymagik @JiggyPete waddayareckon

Probably should add a couple features for wrong email in forgot password form etc.

### DIFF
--- a/spec/features/principal_reset_password_spec.rb
+++ b/spec/features/principal_reset_password_spec.rb
@@ -1,0 +1,108 @@
+RSpec.feature 'Principal can reset their password' do
+  let(:forgot_password_page) { ForgotPasswordPage.new }
+  let(:reset_password_page) { ResetPasswordPage.new }
+  let(:sign_in_page) { SignInPage.new }
+
+  scenario 'Principal can request a password reset' do
+    given_the_principal_user_exists
+    when_they_visit_the_sign_in_page
+    and_they_click_the_forgot_password_link
+    and_they_submit_their_email
+    then_they_see_a_confirmation_message
+    and_get_an_email_with_a_reset_password_link
+  end
+
+  scenario 'Principal can reset password with link' do
+    given_the_principal_user_exists
+    and_they_have_requested_a_password_reset
+    when_they_visit_their_reset_password_link
+    and_they_submit_a_new_password_correctly
+    then_they_see_a_reset_password_confirmation_message
+
+    when_they_sign_out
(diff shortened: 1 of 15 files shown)


spec/features/principal_reset_password_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def and_they_submit_their_email
+    forgot_password_page.email_field.set @user.email
+    forgot_password_page.submit_button.click
+  end
+
+  def then_they_see_a_confirmation_message
+    message = I18n.t('devise.passwords.send_instructions')
+    expect(sign_in_page.flash_message.text).to eq message
+  end
```
munckymagik: I learned we can use `expect(page.thing).to have_text('blah')`. Does that appeals to you?

neoeno: yeah that seems good

spec/support/forgot_password_page.rb
```diff
@@ -0,0 +1,7 @@
+class ForgotPasswordPage < SitePrism::Page
+  set_url '/users/password/new'
+  set_url_matcher %r{/users/password/new}
+
+  element :email_field, '#user_email'
```
munckymagik: Can't use `t-*` style here? (and in the 3 other pages)

spec/features/principal_sign_in_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  def when_they_sign_in_with_incorrect_details
+    visit new_user_session_path
+    sign_in_page.email_field.set @user.email
+    sign_in_page.password_field.set 'not_a_password'
+    sign_in_page.submit_button.click
+  end
+
+  def they_have_logged_in
+    # We don't have access to `current_user` in feature
+    # tests, so we'll rely on this? Until we can test it on content.
```
munckymagik: Should this be a 'TODO'?

spec/support/forgot_password_page.rb
neoeno: I was under the impression that IDs or test classes are acceptable — but I've been a bit lax even in that regard. Will check opinions & fix

benbarnett: Yeah `t-` classes are preferable as it's less likely they'll be used for anything other than the test hook. IDs could be relied on elsewhere (probably js) and things can get messy quite quickly

neoeno:
image

PR #195, opened 2015-05-18
Move delete button away from edit, make both more obvious.

neoeno:
Users rightly pointed out that having the delete and edit buttons close together was perhaps not the best idea. So we've moved the delete button to the right. While we're here, we can make them a little more obvious using the default bootstrap button styles.


benbarnett:
Seems sensible :+1:

paulanthonywilson:
:+1:

PR #196, opened 2015-05-25
Revert rubocop fix as Time.zone is unavailable without Rails.

neoeno:
Also disable Rubocop check, as it's a rails cop and not applicable here.

I wanted to disable all the rails cops for the whole file but couldn't figure out how to do that.


munckymagik:
:+1:

mconnell:
:+1:

PR #197, opened 2015-05-25
Invite existing principals to sign up for a user account.

JiggyPete:
This involved renaming the authentication work that was in place. In the firm area of the site it is based around the principal token. However the current_user actually returned a principle.

First step was to rename current_user to current_principal.

Thereafter devise was introduced
We make use of the sign in/out/forgot password functionality from devise.

Next came devise_invitable. At this point a rake task (notify:existing_firms_to_sign_up) was created to invite all the principals that don't have an account.

The password policy for sign up was that,
* 8 or more characters long
* at least one lowercase letter
* at least one uppercase
* at least numeric character
* at least one symbol
* after 180 days the user must change their password
* previous 12 passwords should be remembered, and not allowed the user to set these again

We made use of devise_security_extension to take care of the password policy.

Finally, new principals need to be able to construct an account without the use of a rake task.
To achieve this we added the password fields to the principal#new stage.


paulanthonywilson:
There's a few commits marked squash/squish/wip. Shouldn't these be squashed?

JiggyPete:
That gets done after the PR is approved

neoeno:
Reformatted this a bit

config/initializers/devise.rb
```diff
@@ -0,0 +1,336 @@
...
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # config.secret_key = 'ddaae8e85b0b01ad44ed021fb9c8b21b8e616bfb41613eefc7df121058cc67347632159785f64502bf89e4e5b2290273b66d0dc58bc026a458bc16e0bcab5c49'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'RADenquiries@moneyadviceservice.org.uk'
```
paulanthonywilson: No biggy, but I'm wondering if this should be in config. Especially as this is a public repo.

paulanthonywilson:
:+1:

darokel:
What's the difference between a user and a principle?

JiggyPete:
A `User` is the object that wraps up all the authentication/account.
A `Principal` is a representation of the person at the firm that is in charge of looking after and administering all of the firms details within RAD.

In one sense they are the same thing, on the other hand it this was to separate out the concerns of authentication.

darokel:
@JiggyPete That makes sense. All godo for this PR :+1:

PR #198, opened 2015-05-26
Add firm index for self-service dashboard

neoeno:
NB: this branch contains a commit marked `[KILL]` that will be removed and replaced by a version of `mas-rad_core` when https://github.com/moneyadviceservice/mas-rad_core/pull/60 is merged.


app/assets/stylesheets/components/firm/_all.scss
```diff
@@ -1,2 +1,3 @@
 @import 'settings';
 @import 'firm';
+@import 'firm_list';
```
benbarnett: Really `firm_list` should be moved into its own directory within components, and have its own settings file.

app/assets/stylesheets/components/firm/_firm_list.scss
```diff
@@ -0,0 +1,27 @@
+$color-firm-list-border-grey: #ccc;
```
benbarnett: This colour can then be moved into settings file mentioned in previous comment

app/assets/stylesheets/components/_navigation.scss
```diff
@@ -1,10 +1,10 @@
-.rad__authentication {
+.rad__navigation {
```
benbarnett: I know this is just a rename but what do you think about losing the `rad__` prefix? I'm unsure what purpose it serves as there doesn't appear to be a `rad` block that it's inheriting from? If it was due to a naming clash then `rad-navigation` would work better.

app/views/shared/_navigation.html.erb
```diff
@@ -0,0 +1,25 @@
+<nav class="t-navigation">
+  <ul class="rad__navigation">
+    <% unless user_signed_in? %>
+      <li class='rad__navigation__item'>
+        <%= link_to t('authentication_sign_in'), new_user_session_path, class: 'rad__navigation__link' %>
```
benbarnett: (If you agree with removing `rad__` then it'll affect these lines too.)

app/assets/stylesheets/components/firm/_all.scss
neoeno: Fair point — didn't think about the directory structure beyond 'this is to do with firms'. But it's about the first BEM component right? Will action.

benbarnett: Yeah exactly, the directory contains the component & any settings. So unless `firm_list` was a child of `firm` (which it isn't) then it deserves its own directory.

app/assets/stylesheets/components/_navigation.scss
neoeno: Ah yes, good point. It was initially due to a naming clash with frontend's `.authentication`, but now we've renamed it we don't have to worry about that anymore.

neoeno:
@benbarnett all actioned

benbarnett:
Nice one. :+1: it is!

JiggyPete:
:+1:

PR #200, opened 2015-06-01
Trading Name Filtering [DEV Approved]

JiggyPete:
Introducing List.js and ListFuzzySearch.js to allow the filtering of a table containing Trading Names (FRN, Trading Name and Principle Name.

We want to have a fuzzy search across all 3 fields on /dashboard/firms.

### DIFF
--- a/app/views/dashboard/firms/index.html.erb
+++ b/app/views/dashboard/firms/index.html.erb
@@ -4,7 +4,7 @@
 <div class="l-constrained">
   <h2><%= t('dashboard.firms_index.firm_heading') %></h2>
   <div class="table-wrapper">
-    <table>
+    <table class="l-half-width">
       <thead>
         <tr>
           <th><%= Firm.human_attribute_name(:fca_number) %></th>
@@ -26,25 +26,28 @@
     </table>
   </div>
 
-  <h2><%= t('dashboard.firms_index.trading_names_heading') %></h2>
-  <div class="table-wrapper">
-    <table>
-      <thead>
-        <tr>
-          <th><%= Firm.human_attribute_name(:fca_number) %></th>
-          <th><%= Firm.human_attribute_name(:registered_name)  %></th>
-          <th><%= t('dashboard.firms_index.principal_column_heading') %></th>
-        </tr>
-      </thead>
-      <tbody>
-        <% @trading_names.each do |trading_name| %>
-          <tr class="t-trading-name-table-row">
-            <td class="t-frn"><%= trading_name.fca_number %></td>
-            <td class="t-firm-name"><%= trading_name.registered_name %></td>
-            <td class="t-principal-name"><%= trading_name.principal.full_name %></td>
+  <div id="firm-list" data-dough-component="FilterTable">
+    <h2><%= t('dashboard.firms_index.trading_names_heading') %></h2>
+    <%= filter_field :trading_names, locale_prefix: 'dashboard.firms_index.trading_names_filter' %>
+    <div class="table-wrapper">
+      <table class="l-half-width">
+        <thead>
+          <tr>
+            <th><%= Firm.human_attribute_name(:fca_number) %></th>
+            <th><%= Firm.human_attribute_name(:registered_name)  %></th>
+            <th><%= t('dashboard.firms_index.principal_column_heading') %></th>
           </tr>
-        <% end %>
-      </tbody>
-    </table>
+        </thead>
+        <tbody class="js-filter-rows">
+          <% @trading_names.each do |trading_name| %>
+            <tr class="t-trading-name-table-row">
+              <td class="js-filterable t-frn"><%= trading_name.fca_number %></td>
+              <td class="js-filterable t-firm-name"><%= trading_name.registered_name %></td>
+              <td class="js-filterable t-principal-name"><%= trading_name.principal.full_name %></td>
+            </tr>
+          <% end %>
+        </tbody>
+      </table>
+    </div>
   </div>
 </div>
--- a/spec/javascripts/fixtures/FilterTable.html
+++ b/spec/javascripts/fixtures/FilterTable.html
@@ -1,0 +1,27 @@
+<div>
+  <div id="any_id" data-dough-component="FilterTable">
+    <input type="text" class="js-filter-field">
+    <table>
+      <thead>
+        <tr>
+          <th>ID</th>
+          <th>Name</th>
+        </tr>
+      </thead>
+      <tbody class="js-filter-rows">
+        <tr>
+          <td>1</td>
+          <td class="js-filterable">Alex</td>
+        </tr>
+        <tr>
+          <td>2</td>
+          <td class="js-filterable">Cameron</td>
+        </tr>
+        <tr>
+          <td>3</td>
+          <td class="js-filterable">Horse</td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+</div>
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/components/firm_list/_firm_list.scss
```diff
@@ -23,3 +23,17 @@
 .firm-list__item__name {
   font-weight: bold;
 }
+
+.firm-list__filter {
+  font-size:  $baseline-unit*2;
+  font-color: $color-firm-list-border-grey;
```
neoeno: `font-color` -> `color`. I'll fix it tho since I have it open!

neoeno:
@JiggyPete Just pushed a few things — treat them as suggestions and feel free to delete the commits if you disagree!

JiggyPete:
@neoeno made a wee tweak to get the UI looking pretty again

neoeno:
Now I think of it, the text in the filter input is a bit low contrast, and a fair bit smaller than text fields in the rest of MAS. Sorry — should have picked that up before.

@benbarnett do we have an existing pattern/style for search/filter fields?

app/assets/javascripts/modules/Filter.js
```diff
@@ -0,0 +1,10 @@
+define(['jquery', 'List', 'ListFuzzySearch'], function($, List, ListFuzzySearch) {
+  'use strict';
+
+  var firmOptions = {
+    valueNames: ['js-frn', 'js-firm-name', 'js-principal-name'],
+    plugins: [ListFuzzySearch()]
+  };
+
+  var firmList = new List('firm-list', firmOptions);
```
munckymagik: `var firmList` is unused no?

munckymagik: And I assume the List object can't get garbage collected because it attaches itself to the DOM? Worth checking cause things won't get garbage collected immediately and different browsers behave differently right?

JiggyPete: Good spot, thanks!

benbarnett:
This PR seems pretty active still, can we take off [DEV Approved]?

munckymagik:
Also what about a wee JS unit test for the Filter.js?

munckymagik:
@benbarnett it was taken off yesterday I think, is it still showing up somewhere?

benbarnett:
@munckymagik sorry yep, just hadn't reloaded – good stuff

app/assets/stylesheets/components/firm_list/_firm_list.scss
```diff
@@ -23,3 +23,19 @@
...
   font-weight: bold;
 }
+
+.firm-list__filter {
+  font-size: $baseline-unit*2;
+  color: $color-firm-list-border-grey;
+  margin-top: 44px;
+  margin-left: 0.5em;
+  padding: 0.5em;
+  -webkit-border-radius: 15px;
```
benbarnett: We use [autoprefixer](https://github.com/moneyadviceservice/frontend/blob/master/Gemfile#L65) on the frontend so we don't need to specify these. Probably worth putting on here too.

app/assets/stylesheets/components/firm_list/_firm_list.scss
```diff
@@ -23,3 +23,19 @@
...
+  -moz-border-radius: 15px;
+  border-radius: 15px;
+  border: 2px solid  $color-firm-list-border-grey;
+}
+
+#firm-list h2 {
```
benbarnett: This goes against our CSS convention, can this be given a proper BEM name or use a layout?

app/views/dashboard/firms/index.html.erb
```diff
@@ -19,25 +19,28 @@
...
-        </tr>
-      </thead>
-      <tbody>
-        <% @trading_names.each do |trading_name| %>
-          <tr class="t-trading-name-table-row">
-            <td class="t-frn"><%= trading_name.fca_number %></td>
-            <td class="t-firm-name"><%= trading_name.registered_name %></td>
-            <td class="t-principal-name"><%= trading_name.principal.full_name %></td>
+  <div id="firm-list">
+    <input class="fuzzy-search firm-list__filter" placeholder="filter firms"/>
```
benbarnett: As discussed on Slack with @neoeno, we can't use `placeholder` to act as a label, it should be for example input text only and considered non-essential.

benbarnett: What's the non-Javascript alternative for this functionality?

app/assets/javascripts/modules/Filter.js
```diff
@@ -0,0 +1,10 @@
+define(['jquery', 'List', 'ListFuzzySearch'], function($, List, ListFuzzySearch) {
```
benbarnett: :+1: for wrapping the 3rd party lib in a dough style module

benbarnett: (Extra points available for having this boot from a data attribute using the `componentLoader` and inherit from `DoughBaseComponent`, though. [Example here](https://github.com/moneyadviceservice/pensions_calculator/blob/master/app/assets/javascripts/pensions_calculator/modules/FastClick.js).)

app/assets/javascripts/modules/Filter.js
JiggyPete: I've got it running locally in Firefox, Chrome and Safari after removing the var \o/

app/views/dashboard/firms/index.html.erb
JiggyPete: It's to help in the situation where a company has a long list of trading names to save them looking through them all. 

There isn't any specific non-JS functionality other than the browsers built in `find` functionality.

benbarnett:
@JiggyPete for completeness, would you mind adding an image of the design you're implementing in the description of the PR?

app/views/dashboard/firms/index.html.erb
benbarnett: Can we implement a server side version? Or worst case, just hide this field for non-js users.

rnalexander:
just testing something please ignore this comment.

neoeno:
OK —

1. Restyled the component.
2. Extracted filter field markup to a helper, including adding label, etc.
3. Worked the existing JS into a dough component called TableFilter.
4. Tested the JS.
5. Rebased everything around a bit (soz)

spec/javascripts/fixtures/FilterTable.html
```diff
@@ -0,0 +1,27 @@
...
+          <td>1</td>
+          <td class="js-filterable">Alex</td>
+        </tr>
+        <tr>
+          <td>2</td>
+          <td class="js-filterable">Cameron</td>
+        </tr>
+        <tr>
+          <td>3</td>
+          <td class="js-filterable">Horse</td>
```
benbarnett: 🐴

benbarnett:
Nailed it. :+1:

munckymagik:
So the issue that's [failed the travis build](https://travis-ci.org/moneyadviceservice/rad/builds/65262780) has been fixed on [another branch](https://github.com/moneyadviceservice/rad/commit/d182e658e4c61ca11f34800299f20dbfeeba5531), which is [already in PR](https://github.com/moneyadviceservice/rad/pull/203).

Happy to say cherry pick it from there to this if you like.

neoeno:
I think I might PR that fix on its own — one sec.

neoeno:
@JiggyPete helpfully noticed that ListJS wants an individual class on each field column (for some reason). Since that's of no relevance to our component, I've let the component handle adding those unique classes and added a test for it.

munckymagik:
:+1:

PR #201, opened 2015-06-01
The basics of breadcrumbs for the dashboard

munckymagik:


tomas-stefano:
:+1: (honestly I disagree with the use of the app/views/shared dir, but since it was already there and the app was using I will give the thumbs) :)

benbarnett:
:+1:

PR #203, opened 2015-06-03
Feature/edit trading names

munckymagik:
Adds the ability to edit the firms trading names.

### DIFF
--- a/app/controllers/dashboard/trading_names_controller.rb
+++ b/app/controllers/dashboard/trading_names_controller.rb
@@ -1,0 +1,52 @@
+module Dashboard
+  class TradingNamesController < ApplicationController
+    before_action :authenticate_user!
+
+    def edit
+      @firm = Firm.find(params[:id])
+      render 'dashboard/firms/edit'
+    end
+
+    def update
+      @firm = Firm.find(params[:id])
+      @firm.update(firm_params) && flash[:notice] = I18n.t('dashboard.firm_edit.saved')
+      render 'dashboard/firms/edit'
+    end
+
+    private
+
+    def principal
+      current_user.principal
+    end
+
+    FIRM_PARAMS = [
+      :email_address,
+      :telephone_number,
+      :address_line_one,
+      :address_line_two,
+      :address_town,
+      :address_county,
+      :address_postcode,
+      :free_initial_meeting,
+      :initial_meeting_duration_id,
+      :minimum_fixed_fee,
+      :retirement_income_products_percent,
+      :pension_transfer_percent,
+      :long_term_care_percent,
+      :equity_release_percent,
+      :inheritance_tax_and_estate_planning_percent,
+      :wills_and_probate_percent,
+      :other_percent,
+      in_person_advice_method_ids: [],
+      other_advice_method_ids: [],
+      initial_advice_fee_structure_ids: [],
+      ongoing_advice_fee_structure_ids: [],
+      allowed_payment_method_ids: [],
+      investment_size_ids: []
+    ]
+
+    def firm_params
+      params.require(:firm).permit(*FIRM_PARAMS)
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


app/controllers/dashboard/trading_names_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      :inheritance_tax_and_estate_planning_percent,
+      :wills_and_probate_percent,
+      :other_percent,
+      in_person_advice_method_ids: [],
+      other_advice_method_ids: [],
+      initial_advice_fee_structure_ids: [],
+      ongoing_advice_fee_structure_ids: [],
+      allowed_payment_method_ids: [],
+      investment_size_ids: []
+    ]
```
tomas-stefano: Is this constant private just for declaration purposes, right? In Ruby we can access this outside of the class. e.g TradingNamesController::FIRM_PARAMS.

tomas-stefano:
:+1:

app/controllers/dashboard/trading_names_controller.rb
munckymagik:
```
irb(main):001:0> class Thing
irb(main):002:1> private
irb(main):003:1> STUFF = [0, 1]
irb(main):004:1> end
=> [0, 1]
irb(main):005:0> Thing::STUFF
=> [0, 1]
irb(main):006:0> Thing::STUFF.pop
=> 1
irb(main):007:0> Thing::STUFF
=> [0]
```

Groan. You are right, private or not makes no difference when what we want is `FIRM_PARAMS = [...].freeze`. Good Ruby-fu there sir.

tomas-stefano: :)

benbarnett: Interesting... can you not have private constants in a class then?

benbarnett: I suppose you'd just use a private method(s)?

munckymagik: Interesting, looks like it is possible if you really want but you have to declare it as private with `private_constant :BLAH`. http://stackoverflow.com/a/11885328/369171

tomas-stefano: private_constant :FIRM_PARAMS

munckymagik:
@tomas-stefano are you happy with that last commit?

app/controllers/dashboard/trading_names_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    before_action :authenticate_user!
+
+    def edit
+      @firm = Firm.find(params[:id])
+      render 'dashboard/firms/edit'
+    end
+
+    def update
+      @firm = Firm.find(params[:id])
+      @firm.update(firm_params) && flash[:notice] = I18n.t('dashboard.firm_edit.saved')
```
mconnell: `&&` ?

munckymagik: Rubocop gave me 2 choices as I'm not allowed an if statement with a one line body.

The important bit is the outcome of the conditional `@firm.update(firm_params)`, I want readers to see this first not last. Which is why I chose not to do `flash[:notice] ..... if @firm.update(firm_params)`, which would be around the wrong way.

Honestly, for the sake of a couple of extra lines I'd much prefer old school `if` statements where you see the conditional first then the outcome indented, but I realise readable code is entirely a subjective thing.

Happy to do whatever here. 
`¯\_(ツ)_/¯`

mconnell: fair enough :+1:

munckymagik:
So I'm going to merge this as I have 2 thumbs and the last change after the first thumb was tiny (I think this complies with the new rules?). If anyone wants anything else changed, I'm happy to action and raise a separate PR.

tomas-stefano:
:+1:

benbarnett:
@munckymagik sounds like a fair way to do things, yep

munckymagik:
@neoeno I need to go. Have rebased on master twice now, can you shepherd this to master for me if travis is ok with it?

PR #204, opened 2015-06-04
Fix brittle test that depends on a randomly ordered array

neoeno:
The fix breaks the dependency on the order of `emails_sent`.


JiggyPete:
:+1:

darokel:
:+1:

PR #205, opened 2015-06-04
Pulling both register links into one place

JiggyPete:
The 'sign up' / 'registration' part of the app is the main flow currently available in RAD.

When pulling in devise I missed one of the 'Sign up' links in the generated views that was routing to a devise/sign_up controller. This PR is to rectify that.


benbarnett:
:+1:

mconnell:
:+1:

PR #206, opened 2015-06-05
enabling fuzzy filtering

JiggyPete:
The list.js component we are using under the hood needs `fuzzy-search` as the class for the input box the user enters their filter text into.

### DIFF
--- a/app/assets/javascripts/modules/FilterTable.js
+++ b/app/assets/javascripts/modules/FilterTable.js
@@ -65,6 +65,18 @@ define(['jquery', 'DoughBaseComponent', 'List', 'ListFuzzySearch'],
     return $.unique(filterClasses);
   };
 
+  /**
+   * enableFuzzySearch
+   *
+   * Find input field with class filterFieldClass then add fuzzy-search class to it.
+   *
+   */
+  FilterTableProto.enableFuzzySearch = function () {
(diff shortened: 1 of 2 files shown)


neoeno:
dang, thought I'd found all the classes ListJS wants. do you mind if we do this through the TableFilter component & add an associated test?

neoeno:
nice :+1:

munckymagik:
:+1:

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -65,6 +65,18 @@ define(['jquery', 'DoughBaseComponent', 'List', 'ListFuzzySearch'],
...
   };
 
+  /**
+   * enableFuzzySearch
+   *
+   * Find input field with class filterFieldClass then add fuzzy-search class to it.
+   *
+   */
+  FilterTableProto.enableFuzzySearch = function () {
+    this.$el.find('.' + this.config.filterFieldClass).each(function(_, cell) {
```
benbarnett: This could be done as a one liner, as in this should do the trick:

```
this.$el.find('.' + this.config.filterFieldClass).addClass('fuzzy-search');
```

neoeno: Good catch actually — they're not cells either. Since this just got merged I'll open a quick PR for it.

PR #207, opened 2015-06-05
Remove FuzzySearch

neoeno:
It's just one thing after another!

So FuzzySearch doesn't search on the text contents of the elements, it searches on the HTML contents of the elements. Additionally, FuzzySearch expects the match to be at the beginning of the string, and ignores any matches after ~40 characters.

Even more additionally, there's a bug in FuzzySearch that ignores all the options we could pass to it to make it behave as we want.

All this together made us think we should trash the fuzzy search and just use regular old search.


benbarnett:
:+1:

mconnell:
:+1:

PR #209, opened 2015-06-08
[DEV APPROVED] Feature to add trading names

neoeno:
There's a list of available trading names that gets imported from the FCA data into Lookup::Subsidiaries. This populates a table, and each row has link that points to a TradingNamesController#new action with a param specifying the FCA record to base it on.

From this point, it's just a regular TradingName, which is a kind of Firm — so we reuse most of the logic for editing firms by inheriting from an abstract controller.


rnalexander:
Starting to look at this now, there's a fair bit in this PR so it might take a little bit.

app/views/dashboard/firms/_form.html.erb
```diff
@@ -0,0 +1,59 @@
...
+  </div>
+<% end %>
+
+<section class="form-section">
+  <div class="form-section__heading-with-button">
+    <% if f.object.trading_name? %>
+      <h2 class="form-section__heading-with-button__heading"><%= t('dashboard.firm_edit.trading_name_heading') %></h2>
+    <% else %>
+      <h2 class="form-section__heading-with-button__heading"><%= t('dashboard.firm_edit.firm_heading') %></h2>
+      <a class="form-section__heading-with-button__button button button--primary"><%= t('dashboard.add_trading_name_button') %></a>
```
benbarnett: I know this was just moved around on this PR but might've missed it before. This classlist got a little crazy.

It might be an idea to use a layout in this context. This would essentially say that all `button`'s in this layout should look have this style. For example:

```css
.l-form-section .button {
  float: right;
}
``` 

Its a cleaner way to override component's initial styles and removes the need to have a big long class list like this one. 

It also avoids components styling the same element and therefore competing for specificity (which I'm guessing is why https://github.com/moneyadviceservice/rad/commit/edf01a33130f37c8da45d394ee7acc91b2f7fdfe might've been there?). With a layout class, you're nearly always one level more 'specific'.

You could just wrap each `form-section` in a `l-form-section` container, or all of them in one – whichever you feel works best.

neoeno: Nesting is allowed?! Dang, why didn't ya tell me. This changes everything. Will implement as you suggest.

I can't remember why the button stuff was there — I think it was a remnant from trying to override something that was coming from a few places. Anyway, we clearly didn't need it in the end!

benbarnett: Nesting multiple components within layouts is allowed, yeah. Sounds good.

neoeno: https://github.com/moneyadviceservice/rad/commit/b9dc8cc6cae531400a19e0fa2e24df25354660a3

Focused on `.form-section__heading-with-button*` as it seemed the only really egregious example.

benbarnett: Much nicer!

benbarnett:
:+1: assuming the only changes are the commits marked [DELETE] or [SQUASH]

rnalexander:
Was there anything we needed on this?

neoeno:
one more thumb!

Gemfile
```diff
@@ -24,7 +24,9 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
-gem 'mas-rad_core', '0.0.54'
+gem 'mas-rad_core',
+    github: 'moneyadviceservice/mas-rad_core',
+    branch: 'feature/add_trading_names'
```
asmega: do you want to switch this back? the branch has been merged

spec/controllers/dashboard/firms_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
+module Dashboard
+  RSpec.describe Dashboard::FirmsController, type: :controller do
+    let(:principal) { FactoryGirl.create(:principal) }
```
tomas-stefano: Most of the places are using just create. It is better to follow this convention.

spec/controllers/dashboard/firms_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    let(:firm) do
+      firm_attrs = FactoryGirl.attributes_for(:firm_with_trading_names, fca_number: principal.fca_number)
```
tomas-stefano: Here too

spec/controllers/dashboard/firms_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      principal.firm.update_attributes(firm_attrs)
+      principal.firm
+    end
+    let(:user) { FactoryGirl.create :user, principal: firm.principal }
```
tomas-stefano: Here too

spec/controllers/dashboard/firms_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        it 'assigns only registered trading names' do
+          get :index
+          expect(assigns(:trading_names).count).to eq 2
+        end
+      end
+    end
+
+    describe 'GET #edit' do
+      let(:trading_name) { FactoryGirl.create(:firm, parent_id: firm.id) }
```
tomas-stefano: here too

spec/controllers/dashboard/firms_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        expect(assigns(:firm)).to eq firm
+      end
+
+      it 'renders the edit page' do
+        expect(response).to render_template 'dashboard/firms/edit'
+      end
+    end
+
+    describe 'PATCH #update' do
+      let(:firm) { FactoryGirl.create(:firm) }
```
tomas-stefano: here too

spec/controllers/dashboard/trading_names_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
+module Dashboard
+  RSpec.describe Dashboard::TradingNamesController, type: :controller do
+    let(:principal) { FactoryGirl.create(:principal) }
```
tomas-stefano: Here too

spec/controllers/dashboard/trading_names_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    let(:firm) do
+      firm_attrs = FactoryGirl.attributes_for(:firm, fca_number: principal.fca_number)
+      principal.firm.update_attributes(firm_attrs)
+      principal.firm
+    end
+    let(:user) { FactoryGirl.create :user, principal: firm.principal }
```
tomas-stefano: Here too

spec/controllers/dashboard/trading_names_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
+module Dashboard
+  RSpec.describe Dashboard::TradingNamesController, type: :controller do
+    let(:principal) { FactoryGirl.create(:principal) }
+    let(:firm) do
+      firm_attrs = FactoryGirl.attributes_for(:firm, fca_number: principal.fca_number)
```
tomas-stefano: Here too

spec/features/dashboard/firms_index_spec.rb
```diff
@@ -16,7 +17,8 @@ def given_i_am_a_fully_registered_principal_user
   end
 
   def and_i_have_a_firm_with_trading_names
-    firm_attrs = FactoryGirl.attributes_for(:firm_with_subsidiaries, fca_number: @principal.fca_number)
+    @lookup_trading_name = FactoryGirl.create(:lookup_subsidiary, fca_number: @principal.fca_number)
+    firm_attrs = FactoryGirl.attributes_for(:firm_with_trading_names, fca_number: @principal.fca_number)
```
tomas-stefano: Here too

Gemfile
neoeno: It has, but not released on RubyGems yet. I usually leave it til the pre-merge squash as things can end up getting raised in review that necessitate further changes to the gem.

asmega: :+1:

spec/controllers/dashboard/firms_controller_spec.rb
paulanthonywilson: We could drop the ``FactoryGirl``, but it would be bad idea.  The ```create``` method isn't very expressive and doesn't show up distinctly on text searches. :-1: on dropping the ```FactoryGirl```.

spec/features/dashboard/firms_index_spec.rb
paulanthonywilson: or not ;-)

rnalexander:
:+1:

spec/controllers/dashboard/firms_controller_spec.rb
neoeno: I'd like us to agree on a standard for this (or agree to have no standard). It's not a good idea to have style conversations repeatedly. 

If it's a blocker for this PR, let's decide it here and store it somewhere that we can refer to. If not, let's discuss it when we're in the office this week (it's dev chat this week I think?)

munckymagik: Keeping the `FactoryGirl.` prefixes in makes it easy to pick out by eye or by search where the fixtures are being created. Inconsistency can be addressed. For me, keeping something that improves readability, even if it's not yet throughout the codebase, is a win.

darokel: I already discussed this with @munckymagik and was happy for `FactoryGirl` to be left in there so it's easy to pick out by eye or by search where the fixtures are being created

darokel:
This is :+1: after clarifying the comments regarding the `FactoryGirl` syntax.

tomas-stefano:
I still disagree. So now you have two ways of doing this. But whatever, :+1:

darokel:
@tomas-stefano I prefer standardising too. We only use the shorter syntax in the `frontend` repo and there aren't that many places within that codebase where we're creating factories. So refactoring the shorter syntax calls to use the more verbose option would be trivial.

neoeno:
All for standardising. It's hard to conform to an unwritten standard though — so if we'd like to pick a standard approach, it should be somewhere everyone can refer to. Ideally we'd discuss it outside of PRs too — takes some of the pressure off. Anyway, thanks for the review everyone!

spec/controllers/dashboard/firms_controller_spec.rb
benlovell: `grep 'create(' spec/**/*_spec.rb`

PR #211, opened 2015-06-09
Sort the list of trading names on the firm index page.

munckymagik:

PR #212, opened 2015-06-09
[WIP] Fix intermittent failure that occurs with specific seed

munckymagik:
Running `bundle exec rspec --seed 43693` causes a repeatable test
failure.

Fix seems to be due to the difference between using the `.first` scope
compared with fetch the first record in an expected result set.

`.first` seems to add additional constraints (e.g. order).

### DIFF
--- a/spec/features/dashboard/trading_names_edit_spec.rb
+++ b/spec/features/dashboard/trading_names_edit_spec.rb
@@ -57,7 +57,7 @@ RSpec.feature 'The dashboard trading name edit page' do
 
   def then_i_see_the_edit_page_for_the_first_trading_name
     expect(trading_name_edit_page).to be_displayed
-    expect(trading_name_edit_page.firm_name).to have_text @principal.firm.trading_names.first.registered_name
+    expect(trading_name_edit_page.firm_name).to have_text @principal.firm.trading_names[0].registered_name
   end
 
   def when_i_change_the_information


neoeno:
I really don't want this to be a problem but it is.

:+1:

urfolomeus:
Do we know _why_ this happens and _why_ using the zero index works?

munckymagik:
Yes. `.first` adds an order by id.

```
[3] pry(main)> Firm.registered.first
  Firm Load (1.4ms)  SELECT  "firms".* FROM "firms" WHERE ("firms"."email_address" IS NOT NULL)  ORDER BY "firms"."id" ASC LIMIT 1
```

But

```
[4] pry(main)> Firm.registered[0]
  Firm Load (28.0ms)  SELECT "firms".* FROM "firms" WHERE ("firms"."email_address" IS NOT NULL)
```

urfolomeus:
Sneaky! Every day is a school day :)

urfolomeus:
FYI `first` is not the `first` you might think it is :)

The `first` being called here is the [finder method first](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-first), which does ordering to ensure database implementation doesn't give differing results.

rnalexander:
Would it be better to fix the first method usage instead?  I can easily see another dev refactoring back from [0] to '.first' for the sake of clairty.

munckymagik:
Hmmm, looking at the link @urfolomeus shared, I'm not sure I can think how to do this. @rnalexander what are you suggesting?

asmega:
strange I put a sleep in and it failed as I thought it may be a race condition. however putting in a pry seem to make the test pass

```
diff --git a/spec/features/dashboard/trading_names_edit_spec.rb b/spec/features/dashboard/trading_names_edit_spec.rb
index 70571ea..720a134 100644
--- a/spec/features/dashboard/trading_names_edit_spec.rb
+++ b/spec/features/dashboard/trading_names_edit_spec.rb
@@ -57,6 +57,8 @@ RSpec.feature 'The dashboard trading name edit page' do

   def then_i_see_the_edit_page_for_the_first_trading_name
     expect(trading_name_edit_page).to be_displayed
+    # sleep(2)
+    require 'pry'; binding.pry
     expect(trading_name_edit_page.firm_name).to have_text @principal.firm.trading_names.first.registered_name
   end
```

neoeno:
Still seeing the above fail — though that pry is triggered on two specs, and it's the second one that fails.

neoeno:
Maybe there's some improvement hiding away in here. Perhaps we should make the order explicit in the implementation and that will solve our problem. @munckymagik weren't you working on this already?

munckymagik:
Sorry folks, on reflection this is a pretty terrible rushy attempt at fixing this. And I'm no longer convinced this fixes it with certainty.

Going to have a fresh look at this when I reach the office.

asmega:
here it is

```
plee@phlee:~/git/rad > bundle exec rspec --seed 43693

Randomized with seed 43693
.........................................................................................
From: /Users/plee/git/rad/spec/features/dashboard/trading_names_edit_spec.rb @ line 61 RSpec::ExampleGroups::TheDashboardTradingNameEditPage#then_i_see_the_edit_page_for_the_first_trading_name:

    58: def then_i_see_the_edit_page_for_the_first_trading_name
    59:   expect(trading_name_edit_page).to be_displayed
    60:   # sleep(2)
 => 61:   require 'pry'; binding.pry
    62:   expect(trading_name_edit_page.firm_name).to have_text @principal.firm.trading_names.first.registered_name
    63: end

[1] pry(#<RSpec::ExampleGroups::TheDashboardTradingNameEditPage>)> exit
.
From: /Users/plee/git/rad/spec/features/dashboard/trading_names_edit_spec.rb @ line 61 RSpec::ExampleGroups::TheDashboardTradingNameEditPage#then_i_see_the_edit_page_for_the_first_trading_name:

    58: def then_i_see_the_edit_page_for_the_first_trading_name
    59:   expect(trading_name_edit_page).to be_displayed
    60:   # sleep(2)
 => 61:   require 'pry'; binding.pry
    62:   expect(trading_name_edit_page.firm_name).to have_text @principal.firm.trading_names.first.registered_name
    63: end

[1] pry(#<RSpec::ExampleGroups::TheDashboardTradingNameEditPage>)> exit
...............................

Finished in 36.71 seconds (files took 2.78 seconds to load)
121 examples, 0 failures

Randomized with seed 43693

plee@phlee:~/git/rad >
```

munckymagik:
Closing this as I believe the real fix belongs on another branch I already started where the order of results is being changed and so it makes sense to make the order explicit in this test too. See PR #211. Thanks.

PR #213, opened 2015-06-10
[WIP] Feature/filter advisers

neoeno:
No merges please — just somewhere I can comment.

### DIFF
--- a/app/assets/javascripts/modules/FilterTable.js
+++ b/app/assets/javascripts/modules/FilterTable.js
@@ -7,15 +7,16 @@
  * See test fixture for sample markup - /spec/js/fixtures/FilterTable.html
  */
 
-define(['jquery', 'DoughBaseComponent', 'List'],
-       function($, DoughBaseComponent, List) {
+define(['jquery', 'jqueryFastLiveFilter', 'DoughBaseComponent', 'List'],
+       function($, jqueryFastLiveFilter, DoughBaseComponent, List) {
   'use strict';
 
   var FilterTableProto,
       defaultConfig = {
-        filterTargetClass: 'js-filterable',
-        filterFieldClass: 'js-filter-field',
-        filterListClass: 'js-filter-rows'
+        filterTargetClass: '.js-filterable',
+        filterFieldClass: '.js-filter-field',
+        filterListClass: '.js-filter-rows',
+        filterWrapperClass: '.js-filter-wrapper',
       };
 
   function FilterTable($el, config) {
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
   };
 
   /**
-   * makeFilterTargetClasses
+   * toggleFilterWrappers function
    *
-   * Find columns with filterTargetClass and apply individual classes
-   * to them, since this is what ListJS wants.
+   * show/hide fitlerWrappers when all filterTargetClasses are hidden
    *
-   * @return {Array} array of classes used
    */
-  FilterTableProto.makeFilterTargetClasses = function () {
-    var filterClasses = [];
+  FilterTableProto.toggleFilterWrappers = function() {
+  	var self = this;
 
-    this.$el.find('.' + this.config.filterTargetClass).each(function(_, cell) {
-      var $cell = $(cell),
-          filterClass = 'js-filterable-' + $cell.index();
-      $cell.addClass(filterClass);
-      filterClasses.push(filterClass);
-    });
+		$(self.config.filterWrapperClass).each( function(index, element ) {
+			self.toggleFilterWrapper( $(element) )
+		});
+	}
 
-    return $.unique(filterClasses);
+  /**
+   * toggleFilterWrapper function
+   *
+   * show/hide fitlerWrapper when all filterTargetClasses are hidden
+   *
+   */
+
+  FilterTableProto.toggleFilterWrapper = function(element) {
+		var list = element.find(this.config.filterListClass)
+		var rows = $(list[0]).find('tr')
+		var hidden_rows = $(rows).filter(function() {
+		  return $(this).css('display') == 'none';
+		});
+
+		if(hidden_rows.length == rows.length) {
+			element.hide();
+		} else {
+			element.show();
+		}
   };
 
   FilterTableProto.bindFilterToElements = function() {
-    var firmOptions = {
-      valueNames: this.makeFilterTargetClasses(),
-      searchClass: this.config.filterFieldClass,
-      listClass: this.config.filterListClass
-    };
+  	var self = this
 
-    new List(this.$el.attr('id'), firmOptions);
+		$(this.config.filterFieldClass).fastLiveFilter(this.config.filterListClass, {
+			selector: self.config.filterTargetClass,
+			callback: function(total) {
+				self.toggleFilterWrappers();
+			}
+		});
   };
 
   return FilterTable;
--- a/app/assets/javascripts/modules/FilterTable.js
+++ b/app/assets/javascripts/modules/FilterTable.js
@@ -3,12 +3,12 @@
  *
  * Requires an element to have a data-dough-component="FilterTable" attribute and an ID.
  *
- * Makes use of ListJS — see listjs.com.
+ * Makes use of jquery.fastLiveFilter.
  * See test fixture for sample markup - /spec/js/fixtures/FilterTable.html
  */
 
-define(['jquery', 'jqueryFastLiveFilter', 'DoughBaseComponent', 'List'],
-       function($, jqueryFastLiveFilter, DoughBaseComponent, List) {
+define(['jquery', 'jqueryFastLiveFilter', 'DoughBaseComponent'],
+       function($, jqueryFastLiveFilter, DoughBaseComponent) {
   'use strict';
 
   var FilterTableProto,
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/modules/FilterTable.js
```diff
@@ -3,19 +3,20 @@
...
  *
- * Makes use of ListJS — see listjs.com.
+ * Makes use of jquery.fastLiveFilter.
  * See test fixture for sample markup - /spec/js/fixtures/FilterTable.html
  */
 
-define(['jquery', 'DoughBaseComponent', 'List'],
-       function($, DoughBaseComponent, List) {
+define(['jquery', 'jqueryFastLiveFilter', 'DoughBaseComponent'],
+       function($, jqueryFastLiveFilter, DoughBaseComponent) {
```
neoeno: Is there a reason we've switched the order here?

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -3,19 +3,20 @@
...
 
   var FilterTableProto,
       defaultConfig = {
-        filterTargetClass: 'js-filterable',
-        filterFieldClass: 'js-filter-field',
-        filterListClass: 'js-filter-rows'
+        filterTargetClass: '.js-filterable',
+        filterFieldClass: '.js-filter-field',
+        filterListClass: '.js-filter-rows',
+        filterWrapperClass: '.js-filter-wrapper',
```
neoeno: Since we've changed these from classes to selectors — can we rename these according to `s/Class/Selector/`?

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
   };
 
   /**
-   * makeFilterTargetClasses
+   * toggleFilterWrappers function
    *
-   * Find columns with filterTargetClass and apply individual classes
-   * to them, since this is what ListJS wants.
+   * show/hide fitlerWrappers when all filterTargetClasses are hidden
```
neoeno: This doc is a little messy — as was the original one I wrote to be honest. We only need the description line I think.

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
...
-    this.$el.find('.' + this.config.filterTargetClass).each(function(_, cell) {
-      var $cell = $(cell),
-          filterClass = 'js-filterable-' + $cell.index();
-      $cell.addClass(filterClass);
-      filterClasses.push(filterClass);
-    });
+		$(self.config.filterWrapperClass).each( function(index, element ) {
+			self.toggleFilterWrapper( $(element) )
+		});
+	}
```
neoeno: We can rename `index` to `_` — which is a convention that means 'this variable isn't going to be used'. We could also rename `element` to be something more indicative of what it is — like we would in an each block in ruby. `filterWrapper` I suppose?

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
...
 
-    return $.unique(filterClasses);
+  /**
+   * toggleFilterWrapper function
+   *
+   * show/hide fitlerWrapper when all filterTargetClasses are hidden
+   *
+   */
```
neoeno: (I'll assume we'll go through all the docs)

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
...
+
+  FilterTableProto.toggleFilterWrapper = function(element) {
+		var list = element.find(this.config.filterListClass)
+		var rows = $(list[0]).find('tr')
+		var hidden_rows = $(rows).filter(function() {
+		  return $(this).css('display') == 'none';
+		});
```
neoeno: There are a few things here:

1. MAS style dictates that we don't have multiple `var`s in a scope. We wrap them all into one var statement, separated with commas.
2. I believe the style is also to use the `$` prefix on variables that contain jQuery elements (took me a while to get this — coming from Angular).
3. I'm not sure, but I don't think we need to do `$(list[0])` — we can use `list.first()`
4. MAS style dictates using camelCase rather than underscores, so `hidden_rows` -> `hiddenRows`.
5. We also don't need that extra jQuery wrapper on `$(rows)`.
6. If you like, I think you could use `$(rows).filter(':hidden')` instead of the filter. That's just an idea though — each approach has its merits.
7. `element` -> `filterWrapper` maybe? There might be a better name than filterWrapper — tableWrapper?

app/assets/javascripts/modules/FilterTable.js
```diff
@@ -45,34 +46,49 @@ define(['jquery', 'DoughBaseComponent', 'List'],
...
   FilterTableProto.bindFilterToElements = function() {
-    var firmOptions = {
-      valueNames: this.makeFilterTargetClasses(),
-      searchClass: this.config.filterFieldClass,
-      listClass: this.config.filterListClass
-    };
+  	var self = this
 
-    new List(this.$el.attr('id'), firmOptions);
+		$(this.config.filterFieldClass).fastLiveFilter(this.config.filterListClass, {
```
neoeno: If we define a `self`, I usually prefer to use `self` throughout — even where you can use `this`. It makes the code a little easier to work with.

app/views/dashboard/advisers/_advisers_table.html.erb
```diff
@@ -1,28 +1,36 @@
...
-    <table class="l-half-width">
-      <thead>
+<div class="table-wrapper js-filter-wrapper">
+  <div class="l-top-right-action">
+    <%= add_adviser_button(firm: firm) %>
+  </div>
+
+	<% if title.present? %>
+		<h2><%= title %></h2>
+	<% end %>
```
neoeno: I'd prefer we keep `_advisers_table.html.erb` just as the table. Apart from anything else — we'd expect `.table-wrapper` to wrap a table, not headers and buttons and such. 

I can see why you've done it — because we need to hide the headers too — but I think we have to solve that via another refactor. If we _do_ need to hide those headings, we need to rename `FilterTable`, since it's no longer just filtering tables but doing other stuff too.

app/views/dashboard/advisers/index.html.erb
```diff
@@ -1,20 +1,14 @@
 <%= render_breadcrumbs [{ locale_key: 'dashboard.navigation.root', url: dashboard_root_path },
                         { locale_key: 'dashboard.navigation.advisers' }] %>
 
-<div class="l-constrained">
-  <div class="l-top-right-action">
-    <%= add_adviser_button(firm: @firm) %>
-  </div>
-
+<div class="l-constrained" data-dough-component="FilterTable" >
```
neoeno: This is more evidence that our component no longer does what its name suggests.

spec/javascripts/tests/FilterTable_spec.js
```diff
@@ -1,33 +1,30 @@
...
   'use strict';
 
-  function fireKeyup(el) {
-    el = el[0] || el; // Unwrap JQuery objects
-    var evObj = document.createEvent('UIEvents');
-    evObj.initUIEvent('keyup', true, true, window, 1);
-    el.dispatchEvent(evObj);
+  function visibleRows() {
+  	return $('.js-filter-rows').find('tr:visible')
+  }
```
neoeno: This is a much better place than getRows previously lived! Not sure where `$` is coming from now, though.

spec/javascripts/tests/FilterTable_spec.js
```diff
@@ -1,33 +1,30 @@
...
+  	$('.js-filter-field').val(text);
+  	$('.js-filter-field').change();
   }
 
   describe('on a table', function() {
     beforeEach(function (done) {
       var self = this;
 
-      requirejs(['jquery', 'FilterTable'], function ($, FilterTable) {
+      requirejs(['jquery', 'jqueryFastLiveFilter', 'FilterTable'], function ($, jqueryFastLiveFilter, FilterTable) {
```
neoeno: Curiosity — why did you add this arg in the middle?

neoeno:
@JiggyPete

Globally applicable things —

1. Whitespace: we need to sort out a lot of rogue tabs here.
2. JS Style: There is a fair amount of inconsistent spacing in function declarations and such, and also there are a bunch of missing semicolons. I don't like semicolons, but I guess we should add them.

Worth saying, after all that — thanks for putting the work in here. JS Plugin DOM Glue is one of my least favourite tasks! But hopefully we'll put it to bed soon.

Also, sorry for making travis fail with my sloppy ruby style!

app/assets/javascripts/modules/FilterTable.js
JiggyPete: keeping the jquery stuff closer together

app/assets/javascripts/modules/FilterTable.js
JiggyPete: Will do

neoeno:
Just opened this up in the browser, noticed a couple things:

image

1. 'Filter trading names' label isn't the right wording anymore. Neither is the placeholder text.
2. The 'Add adviser' button should float to the right of the heading, but it doesn't on the parent firm record.
3. The parent firm heading always sticks around, which could be confusing. Might be worth talking to Spencer about.
4. Not related to this — but we can probably take the `l-half-width` class out, since we can use all that width now.

PR #215, opened 2015-06-12
Add adviser form template

neoeno:
This is one half of a PR we're breaking into two PRs — apologies if any of the code reflects this!

image

### DIFF
--- a/app/controllers/dashboard/advisers_controller.rb
+++ b/app/controllers/dashboard/advisers_controller.rb
@@ -7,10 +7,20 @@ module Dashboard
       @trading_names = @firm.trading_names
     end
 
+    def new
+      @firm = Firm.find(params[:firm_id])
+      @adviser = advisers.build
+    end
+
     private
 
     def principal
       current_user.principal
+    end
+
+    def advisers
+      Firm.find_by(id: params[:firm_id], fca_number: principal.fca_number)
+        .advisers
     end
   end
 end
--- a/app/controllers/lookup/advisers_controller.rb
+++ b/app/controllers/lookup/advisers_controller.rb
@@ -1,12 +1,12 @@
 module Lookup
-  class AdvisersController < PrincipalsBaseController
+  class AdvisersController < ApplicationController
     def show
       @adviser = ::Adviser.find_by(reference_number: params[:id])
 
       if @adviser
         errors_for :conflict
       else
-        @lookup_adviser = Adviser.find_by(reference_number: params[:id])
+        @lookup_adviser = Lookup::Adviser.find_by(reference_number: params[:id])
 
         if @lookup_adviser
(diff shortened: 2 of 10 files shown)


app/controllers/dashboard/advisers_controller.rb
```diff
@@ -7,10 +7,20 @@ def index
...
+
     private
 
     def principal
       current_user.principal
     end
+
+    def advisers
+      Firm.find_by(id: params[:firm_id], fca_number: principal.fca_number)
+        .advisers
```
tomas-stefano: Why not use `principal.firm.find_by(id: params[:firm_id]).advisers` The model has one firm. https://github.com/moneyadviceservice/mas-rad_core/blob/master/app/models/principal.rb#L7

neoeno: That won't work — because `principal.firm` returns a Firm object, not a relation. So I guess you meant: `principal.firm.advisers`?

We don't do that because while the Principal model claims seems to only have one firm, that's a bit misleading. The Principal only has one Firm directly, but that Firm can have its own child Firms. [See here.](https://github.com/moneyadviceservice/mas-rad_core/blob/master/app/models/firm.rb#L34) Each of these will have the same `fca_number`. It took us a while to get our heads around.

So we need to specify the ID to make sure we get the one we're interested in. We need to specify the `fca_number` so the principal can only access their own firms.

tomas-stefano: So, `principal.firm.trading_names.find_by(id: params[:id]).advisers` does the trick?

neoeno: Not if we need the parent firm's advisers (`principal.firm.advisers`)

mconnell:
:+1:

app/controllers/dashboard/advisers_controller.rb
tomas-stefano: Anyway this line is strange and show something wrong with the design. I will loook to this more further after. Given that I will give the thumbs :+1:

neoeno: I completely agree — we've managed to get some refactoring time on this codebase in the next couple of weeks. Your perspective on what you think the problems are and how you'd solve them are definitely appreciated.

neoeno: Thanks for reviewing!

tomas-stefano: This is good thing to present on the fireside chats btw. We will have more eyes to it and is good. Wdyt?

neoeno: That's a fantastic idea. Tagging @darokel so I don't forget to mention this ahead of when we're next in London

PR #218, opened 2015-06-15
On the dashboard, handle the case when no advisers have been added to one or any firms yet

munckymagik:
Adds a helpful message to the table of advisers for each firm, when no advisers have been added yet.

screen shot 2015-06-15 at 14 14 32

## Testing locally

1. Follow the RAD set up instructions in the README.
2. Fire up a Rails console
3. Run the following to create a user/principal who has a firm with trading names but hasn't added any advisers yet:

```ruby
principal = FactoryGirl.create(:principal, fca_number: 999901)
user = FactoryGirl.create(:user, principal: principal)
firm_attrs = FactoryGirl.attributes_for(:firm_with_trading_names, fca_number: principal.fca_number)
principal.firm.update_attributes(firm_attrs)
```

The username is whatever is in `user.email` and the test password (from the Factory) is `Password1!`.

4. Fire up the local Rails development server 
5. Navigate to [http://localhost:3000/dashboard/advisers](http://localhost:3000/dashboard/advisers)

### DIFF
--- a/app/views/dashboard/advisers/_advisers_table.html.erb
+++ b/app/views/dashboard/advisers/_advisers_table.html.erb
@@ -9,20 +9,31 @@
       </tr>
     </thead>
     <tbody>
-      <% advisers.each do |adviser| %>
-        <tr class="t-advisers-table-row">
-          <td class="t-adviser-reference-number"><%= adviser.reference_number %></td>
-          <td class="t-adviser-name">
-            <%= link_to adviser.name, edit_dashboard_firm_adviser_path(adviser.firm, adviser), class: 't-edit-link' %>
+      <% if firm.advisers.present? %>
+        <% firm.advisers.each do |adviser| %>
+          <tr class="t-advisers-table-row">
+            <td class="t-adviser-reference-number"><%= adviser.reference_number %></td>
+            <td class="t-adviser-name">
+              <%= link_to adviser.name, edit_dashboard_firm_adviser_path(adviser.firm, adviser), class: 't-edit-link' %>
+            </td>
+            <td>
+              <ul class="l-comma-separated-list">
+                <% adviser.qualifications.each do |qualification| %>
+                  <li class="t-adviser-qualification"><%= qualification.name %></li>
+                <% end %>
+              </ul>
+            </td>
+            <td class="t-adviser-postcode"><%= adviser.postcode %></td>
+          </tr>
+        <% end %>
+      <% else %>
+        <tr>
+          <td class="t-no-advisers-message" colspan="4">
+            <%= t('dashboard.advisers_index.no_advisers_message') %>
+            <%= link_to(t('dashboard.advisers_index.add_adviser_link', firm_name: firm.registered_name),
+                        new_dashboard_firm_adviser_path(firm),
+                        class: 't-add-adviser-link')  %>
           </td>
-          <td>
-            <ul class="l-comma-separated-list">
-              <% adviser.qualifications.each do |qualification| %>
-                <li class="t-adviser-qualification"><%= qualification.name %></li>
-              <% end %>
-            </ul>
-          </td>
-          <td class="t-adviser-postcode"><%= adviser.postcode %></td>
         </tr>
       <% end %>
     </tbody>
--- a/spec/features/dashboard/advisers_index_spec.rb
+++ b/spec/features/dashboard/advisers_index_spec.rb
@@ -10,14 +10,28 @@ RSpec.feature 'The dashboard adviser list page' do
     then_i_can_see_the_advisers_for_each_trading_name
   end
 
+  scenario 'The principal has not added any advisers yet' do
+    given_i_am_a_fully_registered_principal_user
+    and_i_have_a_firm_with_trading_names_and_no_advisers
+    and_i_am_logged_in
+    when_i_am_on_the_principal_dashboard_advisers_page
+    then_the_parent_firm_section_shows_a_prompt_to_add_an_adviser
+    then_each_trading_name_section_shows_a_prompt_to_add_an_adviser
+  end
+
   def given_i_am_a_fully_registered_principal_user
     @principal = FactoryGirl.create(:principal)
     @user = FactoryGirl.create(:user, principal: @principal)
   end
 
-  def and_i_have_a_firm_with_trading_names_and_advisers
+  def and_i_have_a_firm_with_trading_names_and_no_advisers
     firm_attrs = FactoryGirl.attributes_for(:firm_with_trading_names, fca_number: @principal.fca_number)
     @principal.firm.update_attributes(firm_attrs)
+  end
+
+  def and_i_have_a_firm_with_trading_names_and_advisers
+    and_i_have_a_firm_with_trading_names_and_no_advisers
+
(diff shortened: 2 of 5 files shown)


app/views/dashboard/advisers/_advisers_table.html.erb
```diff
@@ -9,18 +9,29 @@
...
-                <li class="t-adviser-qualification"><%= qualification.name %></li>
-              <% end %>
-            </ul>
+      <% if firm.advisers.present? %>
+        <% firm.advisers.each do |adviser| %>
+          <tr class="t-advisers-table-row">
+            <td class="t-adviser-reference-number"><%= adviser.reference_number %></td>
+            <td class="t-adviser-name"><%= adviser.name %></td>
+            <td>
+              <ul class="l-comma-separated-list">
```
benbarnett: Not sure this is quite right for a layout. Seems like this should be a component?

munckymagik: @benbarnett I see what you mean. @neoeno ?

If this does need to be fixed can we address it in a separate PR dedicated to this change?

neoeno: @benbarnett I can see what you're getting at — do you have a reference for what is and isn't an `l-`? So I can stop trying to model your brain in my head!

We're trashing that code anyway I believe, so no big concern (and it's just a move in this PR).

spec/features/dashboard/advisers_index_spec.rb
```diff
@@ -44,6 +58,24 @@ def then_i_can_see_the_advisers_for_each_trading_name
...
   end
 
+  def then_the_parent_firm_section_is_showing_a_prompt_to_add_an_adviser
+    expect(advisers_index_page).to have_parent_firm
+    expect(advisers_index_page.parent_firm).to have_no_advisers_message(
+      text: I18n.t('dashboard.advisers_index.no_advisers_message'))
+    expect(advisers_index_page.parent_firm).to have_add_adviser_link
+  end
+
+  def then_each_trading_name_section_is_showing_a_prompt_to_add_an_adviser
```
neoeno: Consider:

```
then_the_parent_firm_section_is_showing_a_prompt_to_add_an_adviser
  --> then_the_parent_firm_section_shows_a_prompt_to_add_an_adviser
then_each_trading_name_section_is_showing_a_prompt_to_add_an_adviser
  --> then_each_trading_name_section_shows_a_prompt_to_add_an_adviser
```

Minor thing, I know — active tense is a little easier to read.

app/views/dashboard/advisers/_advisers_table.html.erb
```diff
@@ -9,18 +9,29 @@
...
+            <td class="t-adviser-postcode"><%= adviser.postcode %></td>
+          </tr>
+        <% end %>
+      <% else %>
+        <tr>
+          <td class="t-no-advisers-message" colspan="4">
+            <%= t('dashboard.advisers_index.no_advisers_message') %>
+            <%= link_to(t('dashboard.advisers_index.add_adviser_button_full', firm_name: firm.registered_name),
+                        new_dashboard_firm_adviser_path(firm),
+                        class: 't-add-adviser-link')  %>.
```
neoeno: Sneaky `.` there — should be in a locale.

app/views/dashboard/advisers/_advisers_table.html.erb
benbarnett: Components are individual "things", while layouts style components for specific contexts. As this is just a list, it's a component rather than a 'layout of components'. Does that help?

neoeno: Kind of. The way I thought about it was that `l-comma-separated-list` 'lays out' a set of list items, i.e. it lays them out horizontally rather than vertically.

But what I'm understanding from you is that a layout — actually I still don't understand. I'm finding it hard to reconcile 'layout of components' with things like `l-constrained`.

So I now know that a `ul` of `li`s is not a layout. If I had a `div` of `button`s that I wanted to lay out side-by-side. Would that be a layout or a component?

Not trying to be difficult! I just know it's something that's going to pay off for me to understand, so trying to do so. Previously I just used used `l-` classes for things that do high level laying out on a page, like `l-2col-even` for instance — but this seems like another important use case to understand.

neoeno: Removed in #220

benbarnett: Yeah, its difficult to explain. So a component is just a section of the UI that could be re-used elsewhere. It'll have a 'default' style, but can be modified. In this case, its just a list. The list could be displayed horizontally or vertically, using a modifier class.

This list can then be positioned on the page using a layout, which will typically set up the grid. You'd want to avoid nesting layouts within layouts too much as that's probably a symptom of the components not being styled flexibly enough in the first place. The layout can override specific styles in the component, but too much of this probably indicates the component itself needs working on.

I guess you could think that if you're using a layout just to style an `li` type element, then that's too low level.

Remember, overall, this is a continuously evolving practice and is HARD. Devices, best practices, features are constantly changing which is why we can't give you 'how to use the microwave' style instructions. Hopefully this helps a little bit.

neoeno: Thanks for yr generosity Ben — I'll keep thinking on it.

app/views/dashboard/advisers/_advisers_table.html.erb
munckymagik: If we push the full stop into the locale then I'll need to duplicate the `dashboard.advisers_index.add_adviser_button_full` as surely we don't want this in the button? @neoeno

app/views/dashboard/advisers/_advisers_table.html.erb
munckymagik: @benbarnett if this is being dealt with separately, do you approve of what is here?

benbarnett: Yeah sure, although a bit confused how #220 plays with this?

neoeno: @benbarnett: https://github.com/moneyadviceservice/rad/pull/220/files#diff-324d23ab5d4c99919cced8da980d50a7L1

benbarnett: @neoeno yeah I can see what it's doing, just a bit confused over the sequencing. Don't worry, seems like it's in hand.

benbarnett:
:+1:

app/views/dashboard/advisers/_advisers_table.html.erb
neoeno: Strictly we would want to duplicate it. It's effectively non-internationalised text, and it's reusing a key that doesn't apply to it. This isn't an `add_adviser_button_full`, it's an `add_adviser_link` — that semantic distinction means we have to add the full stop, therefore enough of a semantic distinction to mean it needs another key. 

Imagine you didn't know what the text really was — would you feel confident using the key here?

That said, who cares. It'd be enough to make it explicit with a `+ '.'` in the code, just so it's obvious it's meant to be there.

neoeno:
:+1:

PR #219, opened 2015-06-15
Add advisers to the dashboard

neoeno:
image

The right hand column is the addition here. The left hand column needs addressing as it currently goes on forever.


app/helpers/dashboard/dashboard_helper.rb
```diff
@@ -22,5 +22,18 @@ def add_adviser_button(firm:)
         concat content_tag(:span, sr_label, class: 'visually-hidden')
       end
     end
+
+    def number_of_advisers(firm)
+      all_advisers_for_firm_and_trading_names(firm).size
+    end
+
+    def most_recently_edited_advisers(firm, limit = 3)
```
asmega: personally i prefer an options hash as I think it's more flexible
```
def most_recently_edited_advisers(firm, options = {})
  limit = options[:limit] || 3
end
```

benbarnett:
:+1: for frontend

app/helpers/dashboard/dashboard_helper.rb
neoeno: How do you feel about keyword args?

paulanthonywilson: I think keyword args are the way forward in this case.

```
def most_recently_edited_advisers(firm, limit: 3)
```


Much cleaner. (Aside, I've been using them a lot on the a different project and I was thinking of proposing "Keyword Arg Better Practice" as a Fireside discussion.)

asmega: sounds good

rnalexander:
:+1: for code

darokel:
It feels a little funny that view helpers are calling queries https://github.com/moneyadviceservice/rad/commit/b374664312cdf8c74dc441f3c60098b743227a2c, that's not something they should be concerned with. Ideally, they should just receive variables to use.

neoeno:
Appreciated. It should live on `Firm`

neoeno:
Fair few changes here. Moved the helper methods that were getting advisers onto `Adviser` in `mas-rad_core` and moved the tests etc. Called it through from the controllers.

This uncovered a lack of test coverage — our feature tests for the dashboard didn't include the advisers list. So I added that, then DRY'd it up a little.

So this'll need rethumbing I guess!

darokel:
:+1: :ship:

munckymagik:
Have reviewed the changes since Ryan's thumb, and I think :+1:

PR #221, opened 2015-06-16
[DEV APPROVED] Add filter for advisers

neoeno:
[See here for gif screenshot thing.](https://moneyadviceservice.slack.com/archives/cultivate/p1434470079000022)

There can be a lot of advisers, so users might want to be able to filter them — even across multiple trading names. Unfortunately we had to make some significant changes across libraries, implementation, and markup to do this. We did do it though.


app/assets/javascripts/modules/MultiTableFilter.js
```diff
@@ -0,0 +1,94 @@
...
+  function MultiTableFilter($el, config) {
+    MultiTableFilter.baseConstructor.call(this, $el, config, defaultConfig);
+  }
+
+  /**
+   * Inherit from base module, for shared methods and interface
+   * @type {[type]}
+   * @private
+   */
+  DoughBaseComponent.extend(MultiTableFilter);
```
benbarnett: This is a great example of how we could do better at updating everyone with changes, but components now need to be named else you'll get warnings in the console with the latest version of Dough. Its simple to do: https://github.com/moneyadviceservice/dough/commit/8479406c3a49f5ca8af43a408c2c07b9b3f0c469

app/assets/javascripts/modules/MultiTableFilter.js
```diff
@@ -0,0 +1,94 @@
...
+   */
+
+  MultiTableFilterProto.toggleFilterWrapper = function(filterWrapper) {
+    var $filterWrapper = $(filterWrapper),
+        $list = $filterWrapper.find(this.config.filterListSelector),
+        $rows = $list.first().find('tr'),
+        hiddenRows;
+
+    hiddenRows = $rows.filter(function() {
+      return $(this).css('display') == 'none';
```
benbarnett: Could try the [`:hidden`](https://api.jquery.com/hidden-selector/) filter here – what do you think for readability?

```js
return $(this).is(':hidden');
```

app/assets/javascripts/application.js
```diff
@@ -8,5 +8,5 @@ require(['jquery', 'componentLoader', 'eventsWithPromises'], function ($, compon
 require(['jquery'], function ($) {
   require(['AdviserAjaxCall']);
   require(['DataTransform']);
-  require(['FilterTable']);
+  require(['MultiTableFilter']);
```
benbarnett: Really we should be using the componentLoader for this stuff, but appreciate that the other Javascript isn't written to Dough's interface so not to worry. Perhaps one for the tech debt list, but not high priority.

app/assets/javascripts/modules/MultiTableFilter.js
```diff
@@ -0,0 +1,94 @@
...
+   */
+  MultiTableFilterProto.init = function(initialised) {
+    this.bindFilterToElements();
+    this._initialisedSuccess(initialised);
+    return this;
+  };
+
+  /**
+   * show/hide filterWrappers when all filterTargetSelectores are hidden
+   */
```
benbarnett: Again, not high priority but these comments could do with a cleanup so they fit jsdoc. [DocBlockr](https://packagecontrol.io/packages/DocBlockr) is sublime makes this pretty much instant.

spec/javascripts/fixtures/MultiTableFilter.html
```diff
@@ -1,6 +1,6 @@
 <div>
-  <div id="any_id" data-dough-component="FilterTable">
-    <input type="text" class="js-filter-field">
+  <div id="any_id" data-dough-component="MultiTableFilter" class='js-filter-group'>
```
benbarnett: inconsistent quotes

spec/javascripts/tests/MultiTableFilter_spec.js
```diff
@@ -0,0 +1,88 @@
...
+  function setFilterText(text) {
+    $('.js-filter-input').val(text);
+    $('.js-filter-input').change();
+  };
+
+  describe('on a table', function() {
+    beforeEach(function (done) {
+      var self = this;
+
+      requirejs(['jquery', 'jqueryFastLiveFilter', 'MultiTableFilter'], function ($, jqueryFastLiveFilter, MultiTableFilter) {
```
benbarnett: Do you need to include `jqueryFastLiveFilter`here? It should be pulled in by `MultiTableFilter`'s dep list.

benbarnett:
Looking good! Would be keen to get this in front of DAC too.

app/assets/javascripts/application.js
neoeno: yer not wrong actually — everything works as expected without this line!

app/assets/javascripts/modules/MultiTableFilter.js
```diff
@@ -0,0 +1,94 @@
...
+ * See test fixture for sample markup - /spec/js/fixtures/MultiTableFilter.html
+ */
+
+define(['jquery', 'jqueryFastLiveFilter', 'DoughBaseComponent'],
+       function($, jqueryFastLiveFilter, DoughBaseComponent) {
+  'use strict';
+
+  var MultiTableFilterProto,
+      defaultConfig = {
+        filterTargetSelector: '.js-filterable',
```
benbarnett: Sorry one other thing – we've typically used `data-dough` style attributes for this so its obvious what these things are for when reading the markup. See how the [TabSelector](https://github.com/moneyadviceservice/dough/blob/master/assets/js/components/TabSelector.js#L21-L22) does it.

app/assets/javascripts/modules/MultiTableFilter.js
neoeno: Ah! I've just been guessing the format of these to try to keep them in roughly the same format. Installed the addon so it'll be correct now. May as well go through and make those changes.

neoeno: @JiggyPete @munckymagik fyi ^^^

neoeno: Here's some reference matter: http://usejsdoc.org/

benbarnett: Whilst we're on this – if you pull down Dough and run `gulp` – you'll see some docs outputted in `docs/js` and hopefuly see the point in them.

spec/javascripts/tests/MultiTableFilter_spec.js
neoeno: I think we need it so we can set up the component directly on the element, [on this line](https://github.com/moneyadviceservice/rad/pull/221/files#diff-1c467cf6ef9609b95c5021d49af8cd26R20)? We could use the component loader I guess (not sure how) but that seems like it'd introduce something unrelated to the test scenario. Wyt?

app/assets/javascripts/modules/MultiTableFilter.js
neoeno: Yes yes, that's usually my preference too — the old JS library didn't support that, hence the classes, but no reason we shouldn't change it now.

spec/javascripts/tests/MultiTableFilter_spec.js
benbarnett: I just would've thought that be requiring `MultiTableFilter` itself, it should pull in the dependencies (which is just how requirejs should work). Although I don't actually know without trying it!

neoeno: As I understand it: the dependencies are loaded by the client, but they won't be available in the scope of the function — since requirejs avoids globals and only gives you things when you ask for them to be direct injected for you. 

Either way, the tests fail with that removed — though it's possible that the syntax might need changing in some other way I'm not aware of...

benbarnett: Ah right, that's cool then. Thanks for trying it out.

neoeno: As @munckymagik pointed out, I was looking at the wrong attribute. You're completely right! We don't need `jqueryFastLiveFilter` here.

neoeno: :egg: :arrow_right: :fearful:

benbarnett: Hooray! That array is now one element lighter.

neoeno:
@benbarnett that's everything I think. Note https://github.com/moneyadviceservice/rad/commit/1fbd5103cd805f7705be5211165e13c701755382 and https://github.com/moneyadviceservice/rad/commit/f3766085e03142b8b862506099180ec4878616e2 .

Agree about DAC. This is unlikely to be perfectly accessible in its current form. I'll see if we can get a DAC review on this whole self-service project, but otherwise it might be good to get some review on just this component.

neoeno:
Well one of my commits disappeared. Great. One sec...

neoeno:
There we go.

munckymagik:
:+1:

benbarnett:
Awesome, this could pretty much go straight into Dough now! In fact, we could quite easily do that and get it on Made with Dough too. :+1:

neoeno:
That'd be cool. We need to make sure that doesn't get put off eternally! I'm not sure how to make that happen though... not sure we have any time to put into migrating it atm — do any frontenders?

neoeno:
I've not squashed https://github.com/moneyadviceservice/rad/commit/c560cce3853522371fb576acc5c956310e942098 as it comes the wrong side of a rename which makes it a big hassle.

PR #223, opened 2015-06-17
Add JSHint to default rake task

neoeno:
I also factored out all the puts because, well, I don't know.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,15 +10,25 @@ include FileUtils
 
 KARMA_COMMAND = 'node_modules/.bin/karma'
 
-task :karma do
-  puts
-  puts 'Running JavaScript Karma specs'
-  puts
+def described_task(name, description:)
+  task name do
+    puts
+    puts "Running #{description}"
+    puts
+    yield
+    puts
+  end
+end
 
+described_task :karma, description: 'Javascript Karma specs' do
   fail 'ERROR: karma is not installed' unless File.exist? KARMA_COMMAND
   sh "#{KARMA_COMMAND} start spec/javascripts/karma.conf.js --single-run=true"
+end
 
-  puts
+described_task :jshint, description: 'JSHint' do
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -10,15 +10,25 @@ include FileUtils
...
+end
 
+described_task :karma, description: 'Javascript Karma specs' do
   fail 'ERROR: karma is not installed' unless File.exist? KARMA_COMMAND
   sh "#{KARMA_COMMAND} start spec/javascripts/karma.conf.js --single-run=true"
+end
 
-  puts
+described_task :jshint, description: 'JSHint' do
+  sh 'node_modules/.bin/jshint app/assets/javascripts spec/javascripts' do
```
munckymagik: Can we factor out the binary existence check here too?

neoeno: not sure about factoring it out — but you're right it should definitely be there

benbarnett:
Always wonder where these things should live with Rails projects. Its common to keep scripts like this in the npm package.json (like the [one in Dough](https://github.com/moneyadviceservice/dough/blob/master/package.json#L7)), but then you'd have duplication. Or does one take the canonical list and execute commands using the other? (e.g. rake runs `npm run-script` type commands). Who knows.

Either way, this makes sense for RAD as the rest of these things are in the Rakefile.

:+1:

JiggyPete:
:+1:

Rakefile
munckymagik: Well, factor it in I mean :)

neoeno:
I like `test.sh` as it's flexible and agnostic, but doesn't have as much meaning outside of Go-bound projects...

That last commit work for you, @munckymagik?

munckymagik:
:+1:

neoeno:
I shouldn't have pressed that button

benbarnett:
When autosquash isn't auto enough...

PR #224, opened 2015-06-18
Feature: Force user to sign in after following link in verification email

JiggyPete:
Here we added devise authentication in the appropriate sections of the application to force user sign in.

Also at the end of the flow we allow the user to navigate to the new dashboard functionality.


spec/support/authentication_steps.rb
```diff
@@ -0,0 +1,9 @@
+module AuthenticationSteps
+  def and_i_sign_in(user)
+    sign_in_page = SignInPage.new
+    sign_in_page.load
+    sign_in_page.email_field.set user.email
+    sign_in_page.password_field.set 'Password1!'
```
benbarnett: (Not a rails person – please disregard if irrelevant)... is it best practice to hardcode test passwords like this?

JiggyPete: Hi @benbarnett, as it's only for testing purposes it seems fine as there isn't any risk associated with it.

benbarnett: Apart from if the test data changes it would have to be changed twice? Fair enough though.

darokel:
:+1:

spec/support/authentication_steps.rb
```diff
@@ -0,0 +1,9 @@
+module AuthenticationSteps
+  def and_i_sign_in(user)
```
neoeno: You could use the warden helper `login_as(@user, scope: user)` instead of this. See dashboard_spec.rb:35

It doesn't go through the pages, but since we're not testing those, nor are we trying to do anything tricky like log in or out part way through — I'd suggest we don't need to include it in our feature tests.

Plus, it'll make them faster! :grinning:

neoeno:
From what we've discussed these sound like good sensible changes too. :+1:

neoeno:
looks good (again) :+1:

PR #225, opened 2015-06-18
Let users log in with FRN or email

neoeno:
Sometimes firm principals want to log in with their email, some with their firm reference number.

That's fine, we can do that.


JiggyPete:
:+1:

darokel:
:+1:

PR #226, opened 2015-06-19
Replace firm edit principal section with inline fields and remove groups

neoeno:
## Before

image

## After

image


benbarnett:
:+1:

asmega:
:+1:

PR #227, opened 2015-06-19
[DEV APPROVED] Update the Firm list on the Dashboard to be a most-recently-edited list with up to 3 items

munckymagik:
This updates applies just to the Firm list, the Adviser list will be dealt with in a subsequent PR (there's enough in this already).

It begins implementation of the latest screen designs (there's a quite a bit more to do).

When there are less than 3 firms:

screen shot 2015-06-19 at 12 27 54

When there are more than 3:

screen shot 2015-06-19 at 12 31 21

It depends on https://github.com/moneyadviceservice/mas-rad_core/pull/71 being merged first.


benbarnett:
:+1:

JiggyPete:
:+1:

PR #229, opened 2015-06-19
Fix firm page logic

munckymagik:
A logic error slipped through. This hardens the test by making deeper checks, which teases out the error, then we fix it.


asmega:
:+1:

urfolomeus:
:1:

PR #231, opened 2015-06-19
Fix double flash message

neoeno:
@paulanthonywilson pointed out what I should probably have known — if you set a flash message and then render a view, that flash message will appear twice (first on the page you're rendering, then on the next page when it'll be cleared).

Since we currently render the edit page when we update or create records, we were encountering this. Likewise, if the user hit `#create`, which rendered `:edit`, then hit refresh, unexpected things might happen.

This PR changes some of those renders to redirects, and some of them to use `flash.now`

### DIFF
--- a/app/controllers/self_service/abstract_firms_controller.rb
+++ b/app/controllers/self_service/abstract_firms_controller.rb
@@ -8,8 +8,12 @@ module SelfService
 
     def update
       @firm = Firm.find(params[:id])
-      @firm.update(firm_params) && flash[:notice] = I18n.t('self_service.firm_edit.saved')
-      render :edit
+      if @firm.update(firm_params)
+        flash[:notice] = I18n.t('self_service.firm_edit.saved')
+        redirect_to_edit
+      else
+        render :edit
+      end
     end
 
     protected
@@ -46,6 +50,14 @@ module SelfService
 
     def firm_params
       params.require(:firm).permit(*FIRM_PARAMS)
+    end
+
+    def redirect_to_edit(firm_id: params[:firm_id])
+      redirect_to(
+        controller: params[:controller],
+        action: :edit,
+        firm_id: firm_id
+      )
     end
   end
 end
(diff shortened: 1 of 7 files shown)


urfolomeus:
Are you following a pattern for when you use a redirect and when you use a render?

neoeno:
Uhhh, maybe, but I'd be lying if I said it was notarised or anything. What pattern do you use?

paulanthonywilson:
I’ve always used the “if this request has affected anything, then redirect” pattern, or “if it’s not a GET request, then redirect” which is effectively the same thing. I believe you should be able to refresh your browser without fear of changing anything, or (mostly) getting the “Are you sure you want to resubmit this form” warning.

urfolomeus:
What Paul said :)

munckymagik:
Well, I'd say :+1: to this.

neoeno:
Rebased, still needs work to change some `renders` to `redirect_to`s

neoeno:
Now ready for re-review

app/controllers/self_service/abstract_firms_controller.rb
```diff
@@ -47,5 +51,12 @@ def principal
     def firm_params
       params.require(:firm).permit(*FIRM_PARAMS)
     end
+
+    def redirect_to_edit(firm_id: params[:firm_id])
+      redirect_to url_for(
```
paulanthonywilson: I'm wondering why this is ```url_for``` rather than ```path_for```. Is there not a url helper for this route?

neoeno: No such method as `path_for`. Don't know why...

neoeno: As discussed, dropped the redundant `url_for`

darokel:
:+1:

paulanthonywilson:
:+1:

PR #233, opened 2015-06-23
rename dashboard to selfservice

JiggyPete:
Currently we have a `/dashboard` as the root url namespace, 
all controllers have that namespace, 
`dashboard.en.yml` contains all the translations
and there are a number of scss files relating to this work.

This PR is to rename all of this, so that the routes now look `/selfservice`
and all controllers have the `SelfService` namespace,
translations are now in self_service.en.yml
and scss classes now refer to `self-service`.


app/assets/stylesheets/page_specific/_self_service.scss
```diff
@@ -1,4 +1,4 @@
-.dashboard__action-button {
+.self -service__action-button {
```
neoeno: ;)

config/routes.rb
```diff
@@ -28,7 +28,7 @@
     resources :advisers, only: :show
   end
 
-  namespace :dashboard do
+  namespace :self_service, path: 'selfservice' do
```
neoeno: why the squashed up path?

JiggyPete: I figured that was what MAS would want. It's been updated as per rails default self_service

neoeno:
:+1:

rnalexander:
:+1: :shipit:

PR #235, opened 2015-06-23
Ensure MultiTableFilter works with multiple tables

neoeno:
Previously we used jQuery to search across the whole document. But we want to restrict our searches to the component — so we're doing that now.

Also changes the specs to test against this scenario by default.


munckymagik:
:+1:

benbarnett:
:+1:

darokel:
:+1:

neoeno:
Thanks all!

PR #236, opened 2015-06-23
Remove dashboard

JiggyPete:
The purpose of this PR is to remove this dashboard.

After logging in the user was taken to `dashboard#index`.
They are now being routed to the `self_service/firms#index`

This involved removing the dashboard controller, views and styling.
Removing the Dashboard link in the navigation menu.
Renaming the Dashboard breadcrumb to Home
and some test css class names have been renamed.


spec/support/self_service/firms_index_page.rb
```diff
@@ -1,13 +1,15 @@
 module SelfService
   class FirmsIndexPage < SitePrism::Page
     set_url '/self_service/firms'
-    set_url_matcher %r{/self_service/firms}
+    set_url_matcher %r{/self_service}
```
neoeno: Does this need to be `/self_service/firms`?

neoeno: Actually I can see now that you've changed it explicitly — is there a reason?

JiggyPete: yeah, in the feature tests the user logs in and lands at `/self_service` which now displays the `firms_index_page`

neoeno: Should `/self_service` be a redirect to `/self_service/firms` now then? Rather than having a duplicate endpoint for the same result?

JiggyPete: Redirect added

tomas-stefano:
LGTM :+1:

neoeno:
Looks good :+1:

PR #240, opened 2015-06-24
Allow log-in by FRN only.

neoeno:
Email is no longer allowed, under the logic that the email might change, but the FRN is constant.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -41,7 +41,7 @@ en:
         registered_name: Registered name
         percent_total: Percentage total
       user:
-        login: Email or FRN
+        login: Firm Reference Number
     errors:
       models:
         firm:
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -41,7 +41,7 @@ en:
         registered_name: Registered name
         percent_total: Percentage total
       user:
-        login: Email or FRN
+        login: Firm Reference Number
```
munckymagik: Why not FRN if that is the accepted abbreviation?

neoeno: If we can assume everyone who has to log-in here has that information, we're fine. But my idea was that if someone doesn't know what an FRN is, they've got no other context on this page that they can use to figure it out. Hence the expansion.

munckymagik:
:+1:

darokel:
:ship:

PR #241, opened 2015-06-24
Add 'Remove adviser from directory' buttons to /self_service/firms/id/advisers

JiggyPete:
Allow the deletion of advisers within the self-service area.


app/controllers/self_service/advisers_controller.rb
```diff
@@ -39,6 +39,14 @@ def update
       end
     end
 
+    def destroy
+      adviser = @firm.advisers.find(params[:id])
```
neoeno: Third duplication of this line in this file now — private method?

config/routes.rb
```diff
@@ -34,7 +34,7 @@
     resources :trading_names, only: [:new, :create, :edit, :update]
 
     resources :firms, only: [:index, :edit, :update] do
-      resources :advisers, only: [:new, :create, :edit, :update]
+      resources :advisers, only: [:new, :create, :edit, :update, :destroy]
```
neoeno: Full house. Can we drop the `only:` now?

spec/features/self_service/advisers_index_spec.rb
```diff
@@ -10,6 +10,36 @@
...
+    @adviser_name = advisers_index_page.parent_firm.advisers.first.name.text
+    advisers_index_page.parent_firm.advisers.first.delete_link.click
+  end
+
+  def then_i_am_on_the_advisers_page
+    expect(advisers_index_page).to be_displayed
+  end
+
+  def and_i_can_see_a_success_message
+    expected_message = @adviser_name + ' has been successfully removed.'
```
neoeno: Is there a reason not to pull in the translation here? I'm never sure which way to go on this.

app/views/self_service/advisers/_advisers_table.html.erb
```diff
@@ -11,6 +11,7 @@
         <th class="advisers-table__postcode">
           <%= Adviser.human_attribute_name(:postcode) %>
         </th>
+        <th></th>
```
neoeno: Were you still going to do the column reorder? Or shall I pick it up? 

Don't mind which — though we should have a class to set the width for this `th`. See `/app/stylesheets/components/_firms_table.scss`

config/routes.rb
neoeno: Ah, no `show`. Ignore me

JiggyPete: No :index

neoeno: That too!

app/views/self_service/advisers/_advisers_table.html.erb
```diff
@@ -22,6 +23,9 @@
               <%= link_to adviser.name, edit_self_service_firm_adviser_path(adviser.firm, adviser), class: 't-edit-link' %>
             </td>
             <td class="t-adviser-postcode" data-dough-filterable><%= adviser.postcode %></td>
+            <td>
+              <%= link_to 'Remove from Directory', self_service_firm_adviser_path(firm, adviser), method: :delete, data: {confirm: 'Are you sure?'}, :class => 'button button--small t-remove-button' %>
```
benbarnett: Do we need to use ujs for this? This seems like just a basic onclick listener with a `confirm`?

spec/features/self_service/advisers_index_spec.rb
JiggyPete: I figured we wanted to check the message rather than look it up.

app/views/self_service/advisers/_advisers_table.html.erb
JiggyPete: Not got to it yet.... wasn't sure what you were up to in the background

app/views/self_service/advisers/_advisers_table.html.erb
JiggyPete: There is some jquery schennanigans going on that stops the :delete aspect getting picked up in the frontend. Spent a good while trying to avoid this, however all the advice I can find says to use jquery-ujs.

This is also going to be needed in the next few days when we start deleting firms too.

benbarnett: jQuery shenanigans? Can you explain? We can do all of this with our own Dough-style library without venturing into a different Javascript style and libraries.

spec/features/self_service/advisers_index_spec.rb
neoeno: Yea — I can see arguments for and against both for this one, but not much reason to get into it since we're not localising.

Not relevant but interesting: I just found `I18n.t!` that raises an exception when the translation is missing, which can be useful in tests.

app/views/self_service/advisers/_advisers_table.html.erb
JiggyPete: jquery-ujs is being used for two things here.
* enabling the user to send a HTTP :delete to the server.
* the confirm part.

My confusion today was chasing the first point.

All of the html src produced by rails is fine, however jquery neglects to pick up (or hides) the HTTP_DELETE aspect of the link. This results in a Get request being sent to the server, which confuses the router as it tries to hit a `show` action rather than the `destroy` action.

Jquery-ujs enables (or pays attention to) the `delete` part of the request to make it to the server and can be processed correctly by the rails router.

With regards to the :confirm aspect this usually works with rails straight out of the box, however the jquery-ujs part wasn't enabled in this project yet.

benbarnett: This is very possible using jQuery... http://stackoverflow.com/questions/2153917/how-to-send-a-put-delete-request-in-jquery

All ujs is doing is magically creating a form in the background using Javascript, setting the delete method, and then submitting the form. You can see this in the ujs source here: https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L199

It is simply just not possible to make a 'delete' link without a form. Links do GET requests, and that's it. ujs is just masking this functionality for you, and as a result we're pulling in a 500 line library for no reason.

Seems like you just need a basic onclick listener, see if the result of the `confirm` is truthy and then submit the request using the `delete` method. As a side note, `confirm` is a Javascript feature that has been around since the dark ages, rails just hooks into it.

spec/features/self_service/advisers_index_spec.rb
tomas-stefano: Since we will not use the locales for it, can we use `"#{@adviser.name} has been successfully removed."?

neoeno:
All addressed! Ready for more review

app/views/self_service/advisers/_advisers_table.html.erb
JiggyPete: We've made use of rails `button_to` method which builds a form in the background and created a dough component, `ConfirmableForm` that can be reused later. Thanks for the pointers

darokel:
:+1:

spec/javascripts/tests/ConfirmableForm_spec.js
```diff
@@ -0,0 +1,85 @@
...
+        self.$input = self.$html.find('input');
+        self.$form = self.$html.find('form');
+        self.ConfirmableForm = new ConfirmableForm(self.$input).init();
+        sandbox = sinon.sandbox.create();
+
+        done();
+      }, done);
+    });
+
+    itBehavesLikeAConfirmableForm();
```
benbarnett: 101379-jim-carrey-i-like-it-alot-gif-gflm

benbarnett:
Great stuff :+1:

asmega:
:+1:

PR #245, opened 2015-06-26
Bug fix: order the advisers on the self_service advisers index

munckymagik:
Running RSpec with random seed 53592 flushed this out.

<strike>So, this fix is inconsistent (which is why I have put in the TODO). The ordering should be pushed back into `mas-rad_core` as scope. This is what we've done for the other models. My rationale for doing this quick and dirty fix is:</strike>

<strike>1. I'd prefer the fix to get in sooner rather than later because it causes random test fails. This was avoids the lead time of making a new mas-rad_core release now.</strike>
<strike>2. Other updates are scheduled for mas-rad_core and adding a relevant scope can be batched with these.</strike>

<strike>Do you agree?</strike>

UPDATE: this has now been redone post-update to mas-rad_core.

### DIFF
--- a/app/views/self_service/advisers/_advisers_table.html.erb
+++ b/app/views/self_service/advisers/_advisers_table.html.erb
@@ -15,7 +15,7 @@
       </tr>
     </thead>
     <tbody data-dough-filter-rows>
-      <% advisers.each do |adviser| %>
+      <% advisers.sorted_by_name.each do |adviser| %>
         <tr class="t-advisers-table-row">
           <td class="t-adviser-name" data-dough-filterable>
             <%= link_to adviser.name, edit_self_service_firm_adviser_path(adviser.firm, adviser), class: 't-edit-link' %>
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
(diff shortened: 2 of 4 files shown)


JiggyPete:
:+1:

app/views/self_service/advisers/_advisers_table.html.erb
```diff
@@ -16,7 +16,8 @@
     </thead>
     <tbody data-dough-filter-rows>
       <% if firm.advisers.present? %>
-        <% firm.advisers.each do |adviser| %>
+        <%# TODO: `.order(:name)` needs to be pushed into a scope in mas-rad_core %>
```
tomas-stefano: Can we remove TODO from the source code and add to the Project managment app?

darokel:
When are these other updates for mas-rad_core planned to be integrated?

munckymagik:
@darokel doing the updates now in fact. See https://github.com/moneyadviceservice/mas-rad_core/pull/76 and https://github.com/moneyadviceservice/mas-rad_core/pull/75.

munckymagik:
@darokel @tomas-stefano @JiggyPete This has now been redone after an update to mas-rad_core. Would you mind having a fresh look? Thank you

tomas-stefano:
:+1:

neoeno:
:+1:

JiggyPete:
:+1:

munckymagik:
Thanks all.

PR #246, opened 2015-06-26
Delete trading name

JiggyPete:
Add a button to that allows the deletion of each trading name on the self_service/firms page.

### DIFF
--- a/app/controllers/self_service/firms_controller.rb
+++ b/app/controllers/self_service/firms_controller.rb
@@ -9,6 +9,10 @@ module SelfService
     end
 
     def destroy
+      trading_name = principal.firm.trading_names.registered.find(params[:id])
+      trading_name.destroy
+      flash[:notice] = I18n.t('self_service.firm_destroy.deleted', name: trading_name.registered_name)
+
       redirect_to :back
     end
   end
--- a/app/controllers/self_service/firms_controller.rb
+++ b/app/controllers/self_service/firms_controller.rb
@@ -7,13 +7,5 @@ module SelfService
         @trading_names.map(&:registered_name).include? lookup_name.name
       end
     end
-
-    def destroy
-      trading_name = principal.firm.trading_names.registered.find(params[:id])
-      trading_name.destroy
-      flash[:notice] = I18n.t('self_service.firm_destroy.deleted', name: trading_name.registered_name)
-
-      redirect_to :back
-    end
   end
 end
--- a/app/controllers/self_service/firms_controller.rb
+++ b/app/controllers/self_service/firms_controller.rb
@@ -7,5 +7,9 @@ module SelfService
         @trading_names.map(&:registered_name).include? lookup_name.name
       end
     end
+
+    def destroy
+      redirect_to :back
+    end
   end
 end
--- a/config/locales/self_service.en.yml
+++ b/config/locales/self_service.en.yml
@@ -44,7 +44,7 @@ en:
       available_trading_names_filter_placeholder: e.g. firm name
       new_trading_name_link: Add to directory
       delete_trading_name_button: Remove from Directory
-      delete_trading_name_confirmation: Are you sure you want to remove %{name} from the directory?
+      delete_trading_name_confirmation: Are you sure you want to remove %{name} and all associated advisers from the directory?
 
     firm_edit:
       title: "Edit: %{firm_name}"
--- a/config/locales/self_service.en.yml
+++ b/config/locales/self_service.en.yml
@@ -57,7 +57,7 @@ en:
       address_heading: Address
       questionnaire_heading: Questionnaire Responses
 
-    firm_destroy:
+    trading_name_destroy:
       deleted: "%{name} has been successfully removed."
 
     trading_name_new:
--- a/config/locales/self_service.en.yml
+++ b/config/locales/self_service.en.yml
@@ -46,7 +46,6 @@ en:
       delete_trading_name_button: Remove from Directory
       delete_trading_name_confirmation: Are you sure you want to remove %{name} from the directory?
 
-
     firm_edit:
       title: "Edit: %{firm_name}"
       saved: Saved successfully
@@ -57,6 +56,9 @@ en:
       principal_details_heading: Principal Details
       address_heading: Address
       questionnaire_heading: Questionnaire Responses
+
+    firm_destroy:
+      deleted: "%{name} has been successfully removed."
 
     trading_name_new:
       title: "Add to directory: %{firm_name}"
--- a/config/locales/self_service.en.yml
+++ b/config/locales/self_service.en.yml
@@ -43,6 +43,9 @@ en:
       trading_names_filter_placeholder: e.g. firm name
       available_trading_names_filter_placeholder: e.g. firm name
       new_trading_name_link: Add to directory
+      delete_trading_name_button: Remove from Directory
+      delete_trading_name_confirmation: Are you sure you want to remove %{name} from the directory?
+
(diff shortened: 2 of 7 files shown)


app/controllers/self_service/firms_controller.rb
```diff
@@ -7,5 +7,13 @@ def index
         @trading_names.map(&:registered_name).include? lookup_name.name
       end
     end
+
+    def destroy
+      trading_name = principal.firm.trading_names.registered.find(params[:id])
+      trading_name.destroy
```
munckymagik: Are there any possible failure cases here?

neoeno:
Shouldn't this stuff go through `TradingNamesController` — rather than `FirmsController`?

config/locales/self_service.en.yml
```diff
@@ -43,6 +43,8 @@ en:
       trading_names_filter_placeholder: e.g. firm name
       available_trading_names_filter_placeholder: e.g. firm name
       new_trading_name_link: Add to directory
+      delete_trading_name_button: Remove from Directory
+      delete_trading_name_confirmation: Are you sure you want to remove %{name} from the directory?
```
munckymagik: Deleting a trading name cascades to deleting all the advisers associated with it and thus all the questionnaire data related to them. I feel we should make a big thing about this. "Are you sure you want to delete %{name} AND ALL its associated advisers?" or similar. @spencer @neoeno any opinions on this wording-wise?

spec/features/self_service/firms_index_spec.rb
```diff
@@ -68,6 +80,11 @@ def when_i_am_on_the_principals_firms_index_page
     firms_index_page.load
     expect(firms_index_page).to be_displayed
   end
+  alias_method :and_i_am_on_the_principals_firms_index_page, :when_i_am_on_the_principals_firms_index_page
```
munckymagik: (wow amazing, would have expected Rubocop to complain about no whitespace before the alias. #school_day)

config/locales/self_service.en.yml
neoeno: That sounds good, but without the caps.

app/controllers/self_service/firms_controller.rb
JiggyPete: The principal must exist, and they must have a firm.
The trading names could be empty, then the registered scope would have nothing to operate on which would be fine.

The `.find` would get upset if there was no trading name with that id.
The only way I can see that scenario is if someone if hacking the html src, and it seems reasonable for active record to throw an exception in that situation.

Is there something in particular you're thinking of?

spec/features/self_service/firms_index_spec.rb
```diff
@@ -105,6 +122,26 @@ def and_the_trading_names_section_is_showing_a_prompt_to_add_a_trading_name
     expect(firms_index_page).not_to have_trading_names
   end
 
+  def when_i_delete_the_first_trading_name
+    firms_index_page.trading_names.first.delete_button.click
```
munckymagik: Should this test be running using the `:javascript` Capybara driver? Dunno.

JiggyPete:
@cadem Yeah, it probably should. Seemed easier at the time to drop it on the firm_controller as the user would be on the `firms/index`

spec/features/self_service/firms_index_spec.rb
```diff
@@ -105,6 +122,26 @@ def and_the_trading_names_section_is_showing_a_prompt_to_add_a_trading_name
...
+  end
+
+  def and_i_can_see_a_success_message
+    expected_message = "#{@first_trading_name_registered_name} has been successfully removed."
```
munckymagik: My opinion is we should be using `I18n.t!(...)` consistently in the tests. Seeing as we've decided to localise all the strings, not using them in the tests creates a mixed approach which means every time the wording of a string is tweaked it potentially breaks a test too.

config/locales/self_service.en.yml
JiggyPete: I like that

spec/features/self_service/firms_index_spec.rb
JiggyPete: Good point, why is it working?

neoeno: I think that's actually why we didn't have any trouble with the `confirm` dialogues yesterday! Seems nice that it works without JS

spec/features/self_service/firms_index_spec.rb
JiggyPete: Yeah, there are two sides to this. I tend go with having the message in the test so that it is tested.
If the source changes it's reasonable that the tests will need updated.

However the bang does sway me a little :-)

spec/features/self_service/firms_index_spec.rb
```diff
@@ -105,6 +122,26 @@ def and_the_trading_names_section_is_showing_a_prompt_to_add_a_trading_name
...
+    expect(firms_index_page).to have_flash_message(text: expected_message)
+  end
+
+  def and_i_can_see_the_first_trading_name
+    @trading_name_registered_name_to_delete = firms_index_page.trading_names.first.name.text
```
munckymagik: `.trading_names.first` will be ordered by `:id`. We should make the ordering of `trading_names` match whatever the view is doing so we don't get random test failures in future.

app/controllers/self_service/firms_controller.rb
munckymagik: Does `destroy` have a return value or does it raise an exception on failure?

spec/features/self_service/firms_index_spec.rb
munckymagik: Yeah my view on this is, it's not the actual text we want to test because that could be any number of possible translated versions.

spec/features/self_service/firms_index_spec.rb
JiggyPete: `firm_index_page.trading_names.first` is coming from the view. Funnily enough the implementation earlier caused random test failures, this fixed it.

munckymagik: Doh, yes it's a page object. Ignore me ... carry on ...

spec/features/self_service/firms_index_spec.rb
munckymagik: Yeah, yeah, but, but, we're not ensuring the confirm is happening right? That's the important bit no?

JiggyPete: spot on!

neoeno: personally not too concerned about it given we've got JS tests for the component. It'd be nice to have though, if it's easy

app/controllers/self_service/firms_controller.rb
JiggyPete: `destroy` always returns self even when the record has been removed from the db.

munckymagik: Ok so nothing more to do. Fine.

spec/features/self_service/firms_index_spec.rb
munckymagik: So the other day I accidentally deleted a div tag that had a `data-dough-component=` thing in it. Luckily I spotted it and even more luckily I thought to check the diff so I got back the exact line. It was a broken test that showed me I did this.

munckymagik:
:+1:

JiggyPete:
Well spotted

benbarnett:
:+1:

PR #247, opened 2015-06-29
Self service UX tweaks

neoeno:
Collection of little frontend-y bits in here. Commits should explain, feel free to ask for further explanation if anything's unclear.

### DIFF
--- a/app/assets/stylesheets/layouts/_half_width.scss
+++ b/app/assets/stylesheets/layouts/_half_width.scss
@@ -1,4 +1,8 @@
 .l-half-width {
-  padding-right: $gutter-width/2;
-  width: percentage(1/2);
+  width: percentage(1);
+
+  @include respond-to($mq-l) {
+    padding-right: $gutter-width/2;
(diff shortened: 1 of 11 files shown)


app/assets/stylesheets/layouts/_half_width.scss
```diff
@@ -1,4 +1,8 @@
 .l-half-width {
-  padding-right: $gutter-width/2;
-  width: percentage(1/2);
+  width: 100%;
+
+  @include respond-to($mq-l) {
+    padding-right: $gutter-width/2;
+    width: 50%;
```
neoeno: Was feeling so Monday earlier that I thought the `percentage` function was the dough columns function.

Anyway, it turned out that using `@include column(6);` here would involve either doing some float clearing or adding a row element, so doing it with straight percentages seemed more straightforward.

urfolomeus:
LGTM :+1:

benbarnett:
:+1:

neoeno:
Thanks!

PR #249, opened 2015-06-29
Fix self-service navigation.

neoeno:
I accidentally forgot to check signed out desktop and so it looked odd.
This is cleaner anyway, but does rely on you not needing to hide the
first navigation item.

Perfect solution would be JS or some sort of future CSS but this will do.

### DIFF
--- a/app/assets/stylesheets/components/_navigation.scss
+++ b/app/assets/stylesheets/components/_navigation.scss
@@ -2,26 +2,22 @@
   @extend .unstyled-list;
   text-align: right;
   margin-top: 22px;
+  padding-right: 0.5em;
 }
 
 %navigation-item {
   @include inline-block;
   line-height: 1;
-  border-right: 1px solid $color-primary-nav;
-  padding: 0 $baseline-unit*1.5 0 $baseline-unit*0.7;
+  border-left: 1px solid $color-primary-nav;
+  padding: 0 .25em 0 .5em;
 }
 
 .navigation__item {
   @extend %navigation-item;
 }
 
-.navigation__item--last {
-  @extend %navigation-item;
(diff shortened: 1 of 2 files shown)


benbarnett:
:+1:

app/assets/stylesheets/components/_navigation.scss
```diff
@@ -2,26 +2,22 @@
...
+  padding-right: $baseline-unit;
 }
 
 %navigation-item {
   @include inline-block;
   line-height: 1;
-  border-right: 1px solid $color-primary-nav;
-  padding: 0 $baseline-unit*1.5 0 $baseline-unit*0.7;
+  border-left: 1px solid $color-primary-nav;
+  padding: 0 $baseline-unit*0.7 0 $baseline-unit*1.5;
```
benbarnett: Really we shouldn't be going < 1 $baseline-unit as it defeats its purpose. But this seems like a fairly unique instance.

neoeno: Yeah, I agree. Maybe I can use (r)ems. Baseline is semantically a vertical anyway right?

benbarnett: It ends up as those units on the stylesheets anyway based on the $reponsive flag. It is for vertical rhythm, yeah.

munckymagik:
:+1:

app/assets/stylesheets/components/_navigation.scss
neoeno: Switched it to `em`s in any case. Since they're there to compensate for whitespace hanging around so it makes sense. Thanks!

PR #250, opened 2015-06-30
Configure / to redirect to the self_service root when a user is signed-in

munckymagik:
Had to implement the test as a feature and use Capybara so that sign-in could be faked. See comment in spec header.

### DIFF
--- a/spec/features/root_routing_spec.rb
+++ b/spec/features/root_routing_spec.rb
@@ -1,0 +1,26 @@
+#
+# A routing test implemented as a feature. This is necessary so we can fake
+# logins.
+# See: https://github.com/plataformatec/devise/issues/1670#issuecomment-4060164
+#
+
+RSpec.describe 'GET /', type: :feature do
+  context 'when not logged in' do
+    it 'routes to the root_path' do
+      visit '/'
+      expect(current_path).to eq(root_path)
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


pmcostello:
:+1:

tomas-stefano:
:+1:

spec/features/root_routing_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+
+RSpec.describe 'GET /', type: :feature do
+  context 'when not logged in' do
+    it 'routes to the root_path' do
+      visit '/'
+      expect(current_path).to eq(root_path)
+    end
+  end
+
```
urfolomeus: Finding this slightly confusing. Would you not expect to be on the root page if you visit the root page anyway, regardless of authentication?

urfolomeus: Oh nvm. Helps if I read the implementation :)

munckymagik: Because the authenticated version does a redirect, this is here to confirm that the existing behaviour when not logged in, hasn't changed. Does that make sense? Or have I missed something?

urfolomeus: No it's fine. I just read the specs before the implementation and hadn't realised that you were using conditional routing.

PR #251, opened 2015-06-30
Change self-service menu 'Firms' link to become 'Home'

munckymagik:
Also changed the URL it points at to self_service_root_path.

screen shot 2015-06-30 at 12 52 21

### DIFF
--- a/app/views/shared/_navigation.html.erb
+++ b/app/views/shared/_navigation.html.erb
@@ -9,7 +9,7 @@
       </li>
     <% else %>
       <li class='navigation__item'>
-        <%= link_to t('self_service.navigation.firms'), self_service_firms_path, class: 'navigation__link t-navigation-link' %>
+        <%= link_to t('self_service.navigation.root'), self_service_root_path, class: 'navigation__link t-navigation-link' %>
       </li>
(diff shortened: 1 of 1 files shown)


asmega:
:+1:

benbarnett:
Yet to be decided as to whether this is confusing or not. Would `Directory Home` be clearer to be sure you're not going back to the main MAS site? I guess time will tell.

:+1:

urfolomeus:
:+1:

munckymagik:
@benbarnett interesting point.

To be sure we're on the same page here: this isn't the RAD consumer site, which I agree would be confusing because I think that's meant to feel like it's part of the overall site.

This is in RAD signup application and the link goes to the root of the self-service area only, so maybe the crumb and menu should read "Self Service Home"? Dunno. We have user testing coming up which will probably be the best way to answer this question. Shall we stick with this for now?

benbarnett:
Yeah, I think we will only know the answer from user testing. Perhaps just "Where do you think this link would take you" as a relatively direct question could be part of the agenda.

Sounds cool to stick with it for now, as you say.

munckymagik:
Ok. Thank you.

PR #252, opened 2015-07-01
Change business offering percentage fields to booleans.

neoeno:
See https://github.com/moneyadviceservice/mas-rad_core/pull/77

This is pretty much a straight search-and-replace, and I'd rather avoid cleaning up existing code without good reason — so bear that in mind when reviewing!


config/locales/en.yml
```diff
@@ -315,20 +315,19 @@ en:
         subheading: "Percentage of overall business income taken from the following types of advice:"
```
munckymagik: This string will need to be reworded.

neoeno: You're right, I meant to ask Mark about this. Will do

munckymagik:
:+1:

darokel:
:+1:

darokel:
What's the context behind this change?

munckymagik:
@darokel will answer in Slack for ease

PR #256, opened 2015-07-06
Fca report scripts

neoeno:
I made it work, @munckymagik made it right


munckymagik:
@neoeno I felt like converting the bash script to Ruby for the lols. Could you eyeball that to see if it looks correct?

neoeno:
It looks correct. I find it less clear, but I guess it ought to be in Ruby if possible

darokel:
Could I have a little more business context on how/when this used?

fca_import_scripts/report_invalid_advisers.rb
```diff
@@ -0,0 +1,71 @@
...
+  puts [
+    adviser.reference_number,
+    adviser.clone.reload.name,
+    adviser.firm.registered_name,
+    error,
+    "http://radsignup.moneyadviceservice.org.uk/admin/advisers/#{adviser.id}"
+  ].to_csv
+end
+
+log 'Done'
```
tomas-stefano: I would move this to the fca_import_scripts/lib to follow the convention in there.

munckymagik: @tomas-stefano which bit? This is a front-end script not a module in itself, are you suggesting breaking it out?

munckymagik:
@darokel we have to import data from the FCA at intervals. The status of advisers can change (e.g. from approved to not approved). At this point something needs to be done to work out which records need to be investigated to find out if they should be removed from the directory.

Overall this import process is 90% automated. This change automates a little bit more by making it easy for us to generate a report to show which adviser records need to be investigated. Eventually handling this change *may* get built into the application somehow.

darokel:
@munckymagik Thanks! :+1: for the PR

fca_import_scripts/report_invalid_advisers.rb
tomas-stefano: No. I am suggesting to move this script to the fca_import_scripts/lib folder. There are other scripts in there, so I would move this script file to that folder.

munckymagik: @tomas-stefano sorry there's a misunderstanding here, the intention with our `lib` folder is to home scripts that are just modules or helper scripts that are not to be run directly by a user.

We are keeping the executables in the `fca_import_scripts` folder, at the same level as the README to make it obvious which scripts are front-ends.

Moving this into `lib/` would break this organisation, so we'd really like not to do this.

tomas-stefano: Ok :+1:

tomas-stefano:
:+1:

munckymagik:
Thanks guys

PR #257, opened 2015-07-06
Content changes to both the questionnaire and self-service "add an adviser" forms

munckymagik:
A bunch of content changes. Applied to both the sign up questionnaire and the new self service form.

--
screen shot 2015-07-06 at 15 07 04


darokel:
:+1: Thanks for the screenshot

benbarnett:
:+1:

PR #258, opened 2015-07-06
Style tweaks

neoeno:
'Floats' commit fixes this bug in Safari: 

image

---

'Headings' commit fixes this situation:

image

After:

image

@spencer mentioned a way of doing this using table css, but I couldn't find a way that didn't introduce lots of wrapping divs that I'd rather avoid. This solution assumes buttons are narrower than a certain width (5em in title font) though, so alternative opinions encouraged here.


benbarnett:
:+1:

munckymagik:
:+1:

PR #259, opened 2015-07-06
Enable `pg_stat_statements` in schema.

neoeno:
Something has been adding this for ages, and since it's default on heroku and doesn't seem to have any negative effects let's check it in.


munckymagik:
:+1:

darokel:
:+1:

PR #260, opened 2015-07-07
Lock down `devise_invitable` to prevent users inviting others.

neoeno:
This is accessible via `/users/invitations/new` but not linked anywhere. Still, we should make sure it's impossible. It'd be nice to remove it, but that would involve adding a new override controller which seems a bit heavyweight.

As it is, we let users invite a maximum of 0 users (as the docs suggest) and also modify the flash message display to show the "you've got no invites" flash message if anyone ever tries (they won't).


munckymagik:
Can you link to anything which shows the cost of actually taking out the route so we can compare approaches?

neoeno:
This I guess? http://stackoverflow.com/a/11537795

munckymagik:
Ok so looking at that example I'd say we're better to do exactly as you've done. The route we are trying to disable is protected by authentication, and this way avoids us having to duplicate and respecify routes defined by the `invitable` gem. Therefore, :+1: from me.

tomas-stefano:
:+1:

PR #262, opened 2015-07-07
Amend 'Add trading name' forms to handle pre-existing record.

neoeno:
We've got a situation where there are a number of incomplete records in the database. This causes a number of interesting phenomena. In this instance, we look to see if there's an incomplete copy of our requested firm first and use it if it's there. Otherwise we'd need to delete it, because we'd have a dupe.

Our forms weren't written to expect this, so when the 'new' form was passed an existing record it created an unusual URL. To avoid this, we point the form to the edit action if the record is persisted, and the create action if it's new.

### DIFF
--- a/app/helpers/self_service/self_service_helper.rb
+++ b/app/helpers/self_service/self_service_helper.rb
@@ -19,5 +19,10 @@ module SelfService
         concat content_tag(:span, sr_label, class: 'visually-hidden')
       end
     end
+
+    def create_or_update_self_service_trading_names_path(firm)
+      return self_service_trading_name_path(firm) if firm.persisted?
+      self_service_trading_names_path
+    end
   end
 end
--- a/app/views/self_service/trading_names/new.html.erb
+++ b/app/views/self_service/trading_names/new.html.erb
@@ -7,7 +7,7 @@
   </h1>
 
   <%= form_for @firm,
-               url: self_service_trading_names_path(@firm),
+               url: create_or_update_self_service_trading_names_path(@firm),
                builder: Dough::Forms::Builders::Validation,
                html: { class: 'form' } do |f| %>
 
--- a/spec/helpers/self_service/self_service_helper_spec.rb
+++ b/spec/helpers/self_service/self_service_helper_spec.rb
@@ -1,0 +1,23 @@
+module SelfService
+  RSpec.describe SelfServiceHelper, type: :helper do
+    describe '#add_or_edit_self_service_trading_names_path' do
+      subject { helper.create_or_update_self_service_trading_names_path(trading_name) }
+
+      context 'the trading name is an existing record' do
+        let(:trading_name) { FactoryGirl.create(:trading_name) }
+
+        it 'returns the create path' do
+          expect(subject).to eq self_service_trading_name_path(trading_name)
+        end
+      end
+
+      context 'the trading name is an existing record' do
+        let(:trading_name) { FactoryGirl.build(:trading_name) }
+
+        it 'returns the create path' do
+          expect(subject).to eq self_service_trading_names_path
+        end
+      end
+    end
+  end
+end


munckymagik:
:+1:

asmega:
is this what you were after?

```
plee@phlee:~/git/rad > RAILS_ENV=test be rails c
Loading test environment (Rails 4.2.1)
[1] pry(main)> app.url_for [:self_service, Firm.first]
  Firm Load (3.8ms)  SELECT  "firms".* FROM "firms"  ORDER BY "firms"."id" ASC LIMIT 1
=> "http://www.example.com/self_service/firms/5"
[2] pry(main)> app.url_for [:self_service, Firm.new]
=> "http://www.example.com/self_service/firms"
```

neoeno:
It's certainly a nice idea — in this instance we're dealing with a bit of a subtle distinction between firms and trading names. They've been implemented as a sort of unofficial Single Table Inheritance, but the key is that the URLs we want are:

```ruby
# pseudocode:

given existing_firm.trading_name? == true
f(existing_firm) = "http://www.example.com/self_service/trading_names/5"

given new_firm = Firm.new(is_a_trading_name: true)
f(new_firm) = "http://www.example.com/self_service/trading_names"
```

Is there a way of specifying the controller in the scheme you've outlined?

asmega:
```
[17] pry(main)> app.url_for [:self_service, :trading_names, Firm.new]
NoMethodError: undefined method `self_service_trading_names_firms_url' for #<ActionDispatch::Integration::Session:0x007f9423f09e70>
from /Users/plee/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/actionpack-4.2.1/lib/action_dispatch/testing/assertions/routing.rb:171:in `method_missing'
```
maybe then defined this named route in `config/routes.rb`

if not then we can go with what you've implemented

neoeno:
I like the idea of ditching this helper in favour of the routes, but I'm not sure how to implement that. Here's what we've got right now:

```ruby
resources :trading_names, only: [:new, :create, :edit, :update, :destroy]
```

I tried this to see what I have to work with:

```ruby
resources :trading_names, only: [:new, :create, :edit, :update, :destroy] do
    collection do
      post :create, as: :horse
    end
  end
```

And I got:

```bash
$ rake routes | grep horse
           horse_self_service_trading_names POST   /self_service/trading_names(.:format)                                          self_service/trading_names#create
```

I tried a few other combos but couldn't get anything close to what we need. Any ideas?

asmega:
try this

```
plee@phlee:~/git/rad > git diff
diff --git a/config/routes.rb b/config/routes.rb
index 33696b3..2d1038f 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -39,6 +39,9 @@ Rails.application.routes.draw do
   namespace :self_service do
     root to: redirect('/self_service/firms')
     resources :trading_names, only: [:new, :create, :edit, :update, :destroy]
+
+    resources :trading_names, only: [:new, :create, :edit, :update, :destroy], as: 'trading_names_firms'
+
     resources :principals, only: [:edit, :update]

     resources :firms, only: [:index, :edit, :update] do
plee@phlee:~/git/rad >
```
```
[29] pry(main)> app.url_for [:self_service, :trading_names, Firm.first]
  Firm Load (0.5ms)  SELECT  "firms".* FROM "firms"  ORDER BY "firms"."id" ASC LIMIT 1
=> "http://www.example.com/self_service/trading_names/5"
[30] pry(main)> app.url_for [:self_service, :trading_names, Firm.new]
=> "http://www.example.com/self_service/trading_names"
```

you will have to have some dupe routes due to the name. up to you if you prefer this (extra route) or yours extra code

neoeno:
wyt @munckymagik?

munckymagik:
Thanks for the suggestion Phil.

IMO if I was going through the code trying to work out what's going on, seeing the helper and the logic would tell me a lot more about what's going on. So for that reason, my vote is with the helper.

asmega:
ok

PR #266, opened 2015-07-14
Fix invalid principal submission causing 500 error.

neoeno:
The edit view expects `@firm` to be set — but the update action didn't set this when it was rendering the edit view in the case of an invalid submission. We'll move it to a before_action and add an integration test that would have caught this.

Questioning the whole instance variables to pass values to views thing at the moment — but not about to change it here.


darokel:
:+1:

asmega:
:+1:

PR #267, opened 2015-07-15
Fix layout bug in legend tags in IE

neoeno:
The legend tags overflowed their containers and ran amok. This handy width property sorts it all out.

Legend tags seem to have unexpected behaviour in IE (even 11). We might use them with care in future.

Before:
image

After:
image


benbarnett:
:+1:

munckymagik:
:+1:

PR #268, opened 2015-07-16
change rake task to produce csv list of principals to invite

JiggyPete:
This change is to update/replace the existing rake task which would previously invite firms without an account in RAD to create one.

This process is changing so that MailChimp (or similar service) can be used for sending and tracking the invitations. 

The PR is to change the rake task so that it now returns CSV output that can easily be used with MailChimp.

### DIFF
--- a/lib/tasks/export_existing_firms_without_accounts.rake
+++ b/lib/tasks/export_existing_firms_without_accounts.rake
@@ -1,0 +1,6 @@
+namespace :principals do
+  desc 'existing firms needing notified to sign up'
+  task generate_invitations_csv: :environment do
+    Tasks::ExistingFirmsSignUpTask.notify(Tasks::UserInvitationHelper.new)
+  end
+end
--- a/spec/lib/tasks/existing_firms_sign_up_task_spec.rb
+++ b/spec/lib/tasks/existing_firms_sign_up_task_spec.rb
@@ -1,107 +1,109 @@
+class StubInvitationHelper
+  def invitation_url(user)
+    "invitation for user #{user.id}"
+  end
+end
+
 module Tasks
   RSpec.describe ExistingFirmsSignUpTask do
     describe '#notify' do
-      let!(:bills_firm) { create :firm }
-      let!(:bill) { create :principal, email_address: 'bill@example.com', firm: bills_firm }
+      let!(:firm) { create :firm, registered_name: 'Wright, Johnston & MacKenzie LLP' }
+      let!(:principal) { create :principal, firm: firm, first_name: 'Bill', last_name: 'Junior, the third' }
+      let(:stub_inviter) { StubInvitationHelper.new }
+      let(:output) { [] }
 
-      before do
+      it 'invites principals that have no account' do
+        expect(User.count).to eq(0)
+        described_class.notify(stub_inviter, output)
+        expect(User.first).to be_invited_to_sign_up
+      end
+
+      it 'sets user.invitation_sent_at' do
+        expected_invitation_sent_at = DateTime.current
+        allow(DateTime).to receive(:current).and_return(expected_invitation_sent_at)
+
+        described_class.notify(stub_inviter, output)
+        expect(User.first.invitation_sent_at.to_datetime).to be_within(1.second).of(expected_invitation_sent_at)
+      end
+
+      it 'invites but does not send email' do
         ActionMailer::Base.deliveries = []
+        described_class.notify(stub_inviter, output)
+        expect(ActionMailer::Base.deliveries).to be_empty
       end
 
-      it 'mails principals that have no account' do
-        create :principal, email_address: 'ben@example.com', firm: create(:firm)
+      describe 'when creating a csv entry for a principal' do
+        before do
+          described_class.notify(stub_inviter, output)
+        end
 
-        described_class.notify
+        it 'contains the frn' do
+          CSV.parse(output.first) do |line_data|
+            expect(line_data[0].to_i).to eq(principal.firm.fca_number)
+          end
+        end
 
-        expect(emails_sent.size).to eq(2)
-        expect(emails_sent).to all(have_attributes(subject: 'Invitation instructions'))
+        it 'contains the firm name' do
+          CSV.parse(output.first) do |line_data|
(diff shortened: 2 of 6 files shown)


lib/tasks/export_existing_firms_without_accounts.rake
```diff
@@ -0,0 +1,6 @@
+namespace :export do
```
neoeno: Namespaces probably should be nouns rather than verbs, and the task name is a little confusing for me.

How about `principals:generate_invitations` — or something

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
```diff
@@ -1,107 +1,117 @@
...
+      let!(:bills_firm) { create :firm, registered_name: 'Wright, Johnston & MacKenzie LLP' }
+      let!(:bill) do
+        principal = build :principal
+        principal.email_address =  'bill@example.com'
+        principal.firm = bills_firm
+        principal.first_name = 'Bill'
+        principal.last_name = 'Junior, the third'
+        principal.save!
+        principal
+      end
```
neoeno: calling things according to class, like `:principal` rather than `:bill` seems a nice convention to keep to. Also — is there a reason we're not using a factory for this?

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
```diff
@@ -1,107 +1,117 @@
...
-      before do
+        described_class.notify(stub_inviter, output)
+        expect(User.first.invitation_sent_at.to_datetime).to be_within(1.second).of(expected_invitation_sent_at)
+      end
+
+      it 'invites but does not send email' do
         ActionMailer::Base.deliveries = []
+        described_class.notify(stub_inviter, output)
+        expect(ActionMailer::Base.deliveries).to be_empty
       end
```
neoeno: Good thing to test!

lib/tasks/export_existing_firms_without_accounts.rake
JiggyPete: not fussed either way

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
```diff
@@ -1,107 +1,117 @@
...
 
-      it 'mails principals that have no account' do
-        create :principal, email_address: 'ben@example.com', firm: create(:firm)
+      context 'creates a csv entry for principal' do
+        before do
+          described_class.notify(stub_inviter, output)
+        end
 
-        described_class.notify
+        it 'frn' do
```
neoeno: The `it` blocks here and the containing `context` block aren't named according to their conventions. See here for an example for describing `context` blocks: http://betterspecs.org/#contexts (generally 'when' or 'with' are the keywords to look for — though you could rephrase this example in a few ways using `describe` too)

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
JiggyPete: Previously in some of the tests there were two principals. I'm not keen on `principal_1` and `principal_2` as names so `bill` gave this some more context. Happy to change it.

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
```diff
@@ -1,107 +1,117 @@
...
 
-        expect(emails_sent.first.to).to include('firm@example.com')
+        expect(output.length).to eq(1)
+        CSV.parse(output.first) do |line_data|
+          expect(line_data[3]).to eq('firm@example.com')
+        end
       end
 
-      it 'does not mail principals thats firm is not registered' do
+      it 'output does not include details where the firm is not registered' do
```
neoeno: Rephrase to `does not create output including details where the firm is not registered` — in line with the previous `it` block

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
neoeno: For me, when I read it, `bill` gives less context as I'm not sure what kind of a thing `bill` is. Since we're only using one principal here(?) `principal` seems like an adequate name to me.

lib/tasks/export_existing_firms_without_accounts.rake
JiggyPete: how about `principals:generate_csv_invitations` ?

neoeno: seems good, or `principals:generate_invitations_csv`. either is fine

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
neoeno: Also — any reason not to use the factory to set & save these attributes?

lib/tasks/export_existing_firms_without_accounts.rake
JiggyPete: :+1:

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
JiggyPete: Yeah, I didn't clean up. Most of the info was an overhang from the previous implementation using a factory. I added a real value from prod for the `last_name` line so that we checked that commas were escaped in the csv output. 

All of this made rubocop unhappy with the line length or splitting and it ended up as above when I should have cleaned it up with less.

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
JiggyPete: how does this look
`context 'when creating a csv entry for a principal' do`
        ...
     ` it 'contains the frn' do`

spec/lib/tasks/existing_firms_sign_up_task_spec.rb
JiggyPete: Fixed up a few more around this. Appreciating the focus on the language around testing :+1:

rnalexander:
:+1:

neoeno:
:+1:

PR #271, opened 2015-07-17
add column to csv invitation listing

JiggyPete:
We need to be able to split the csv list of data into firms that are registered and those that are not.
The reason for this is to apply one mail template to the registered firms and another to the firms that are not registered.

In this PR we are adding a column in the output csv that will have the value 'registered' or 'not registered'.

### DIFF
--- a/lib/tasks/existing_firms_sign_up_task.rb
+++ b/lib/tasks/existing_firms_sign_up_task.rb
@@ -1,24 +1,16 @@
 module Tasks
   class ExistingFirmsSignUpTask
     def self.notify(inviter, output = STDOUT)
-      valid_registered_parent_firms.each do |firm|
-        principal = firm.principal
-        user = User.find_by_principal_token principal.token
-        next if user.present?
-
-        user = invite principal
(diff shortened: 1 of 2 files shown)


lib/tasks/existing_firms_sign_up_task.rb
```diff
@@ -30,8 +22,17 @@ def self.invite(principal)
...
+      principal = user.principal
+      firm = principal.firm
+
+      result = [firm.fca_number]
+      result << firm.registered_name
+      result << principal.full_name
+      result << principal.email_address
+      result << inviter.invitation_url(user)
+      result << (firm.email_address.present? ? 'registered' : 'not registered')
+      result
```
neoeno: This could be put more concisely like:

```ruby
[
  firm.fca_number,
  firm.registered_name,
  principal.full_name,
  principal.email_address,
  inviter.invitation_url(user),
  (firm.email_address.present? ? 'registered' : 'not registered')
]
```

your call

neoeno:
:+1:

neoeno:
:+1:

darokel:
:+1:

PR #272, opened 2015-07-20
Add email to devise parameter sanitizer

neoeno:
We need this so changing it on the 'change password' page works correctly. We didn't think we needed it because our login form submits the `login` parameter, not `email`.


JiggyPete:
:+1:

benbarnett:
:+1:

PR #273, opened 2015-07-21
Add `required_asterisk` method to add nbsp and * to required fields

neoeno:
Why not use `&nbsp;` you ask? Because I'd have to `#html_safe` the whole string and I prefer to avoid that if possible – even though at the moment there's no interpolation.


benbarnett:
:+1:

darokel:
:+1:

PR #275, opened 2015-07-22
Add new conditional flow for remote & face to face advice method

neoeno:
There's a fair amount of code changed here — apologies! The commits should make sense at least (I hope).

## Before

image

Two sets of checkboxes, one for face to face advice, one for remote advice. Firms could select any combination.

## After

One set of radio buttons to pick a primary advice type, then:

If they pick remote, then they see

* A set of checkboxes to select remote advice methods.

If they pick face-to-face, they see:

* A set of checkboxes to select face-to-face advice methods.
* Another set of checkboxes to select remote advice methods (optional).

---

In terms of the JS, I decided to extend the existing FieldToggleVisibility component to be a bit more robust about letting me show and hide different things. This went pretty well until I needed to handle the situation when neither radio button was selected, and then it didn't work anymore. I added a little bit of extra jQuery to handle that situation — it's not nice but I'm hoping it's adequate. Do you think so?

We'll be working with this page a fair bit in coming months, so maybe I'll get to add in a data binding library for these slightly more involved scenarios.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
-gem 'mas-rad_core', '0.0.63'
+gem 'mas-rad_core', '0.0.64'
 gem 'oga'
 gem 'pg'
 gem 'ransack'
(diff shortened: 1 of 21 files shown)


JiggyPete:
:+1:

tomas-stefano:
:+1:

PR #278, opened 2015-07-23
Allow Admin users to sign-in/behave like Principals

JiggyPete:
To provide good customer service, 
Admin users need to be able to perform any operation within Self-Service that principals can.

This PR provides a button in the admin area of RAD to 'Sign in as <principal>' to allow the admin to edit the principals firm details, without having knowledge of the principals password.

This functionality is only accessible in the admin area of the site.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,0 +1,8 @@
+class Admin::UsersController < Admin::ApplicationController
+  def switch_user
+    user = User.find params[:user_id]
+    sign_in(:user, user)
+    flash[:notice] = "You are now signed in as #{user.principal.full_name}"
+    redirect_to self_service_root_path
+  end
+end
--- a/app/helpers/admin/firms_helper.rb
+++ b/app/helpers/admin/firms_helper.rb
@@ -16,6 +16,10 @@ module Admin::FirmsHelper
     end
   end
 
+  def user_for(principal)
+    User.find_by_principal_token principal.token
+  end
+
   private
 
   def render_literal_or_fee_or_percentage(value, attribute_name)
--- a/spec/features/admin/sign_in_as_principal_spec.rb
+++ b/spec/features/admin/sign_in_as_principal_spec.rb
@@ -1,0 +1,54 @@
+RSpec.feature 'Move advisers between firms' do
+  let(:admin_firm_page) { Admin::FirmPage.new }
+  let(:firms_index_page) { SelfService::FirmsIndexPage.new }
+  let(:updated_postcode) { 'EH11 2AB' }
+
+  scenario 'admin signs in as principal' do
+    given_there_is_a_fully_registered_principal_user
+    and_i_am_logged_in_as_an_admin_user
+    and_i_am_on_the_admin_firm_page
+    when_i_sign_in_as_principal
+    then_i_see "You are now signed in as #{@principal.full_name}"
+    and_i_am_on_the_self_service_firm_page
+  end
+
+  scenario 'admin visits page of firm with a principal without a user record' do
+    # i.e. when the invite rake task hasn't been run
+    given_there_is_a_principal
+    and_i_am_logged_in_as_an_admin_user
+    and_i_am_on_the_admin_firm_page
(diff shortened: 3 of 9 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Admin::UsersController < Admin::ApplicationController
+  def switch_user
+    user = User.find params[:user_id]
+    sign_in(:user, user)
+    flash[:notice] = "You are now signed in as #{user.principal.full_name}"
```
neoeno: Could internationalise this string

neoeno: Actually, the admin stuff isn't internationalized at all is it? In which case leave it I guess!

app/helpers/admin/firms_helper.rb
```diff
@@ -16,6 +16,10 @@ def render_questionnaire_response(firm, attribute_name)
     end
   end
 
+  def user_for(principal)
+    User.find_by_principal_token principal.token
+  end
```
neoeno: Ah I understand what you were talking about earlier now — principal has no method on it to return the user.

app/controllers/admin/users_controller.rb
JiggyPete: At a quick glance through I can't find anything in the admin section that is internationalised.

spec/features/admin/sign_in_as_principal_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    then_i_see "You are now signed in as #{@principal.full_name}"
+    and_i_am_on_the_self_service_firm_page
+  end
+
+  def given_there_is_a_fully_registered_principal_user
+    @principal = FactoryGirl.create(:principal)
+    @user = FactoryGirl.create(:user, principal: @principal)
+  end
+
+  def and_they_have_a_firm_with_an_adviser
```
neoeno: could drop this method — it's not required for anything we're testing against

spec/features/admin/sign_in_as_principal_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    and_i_am_logged_in_as_an_admin_user
+    and_i_am_on_the_admin_firm_page
+    when_i_sign_in_as_principal
+    then_i_see "You are now signed in as #{@principal.full_name}"
+    and_i_am_on_the_self_service_firm_page
+  end
+
+  def given_there_is_a_fully_registered_principal_user
+    @principal = FactoryGirl.create(:principal)
+    @user = FactoryGirl.create(:user, principal: @principal)
```
neoeno: Could replace this with 

```ruby
@user = FactoryGirl.create(:user)
@principal = @user.principal
```

spec/features/admin/sign_in_as_principal_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    admin_firm_page.sign_in_as_principal.click
+  end
+
+  def then_i_see(message)
+    expect(page.text).to include(message)
+  end
+
+  def and_i_am_on_the_self_service_firm_page
+    expect(firms_index_page).to be_displayed
+  end
```
neoeno: Also, got a failin test for ya: 

```ruby
  scenario 'admin signs in as a principal without a user record' do
    # i.e. when the invite rake task hasn't been run
    given_there_is_a_principal
    and_i_am_logged_in_as_an_admin_user
    and_i_am_on_the_admin_firm_page
    then_it_doesnt_blow_up
  end

  def given_there_is_a_principal
    @principal = FactoryGirl.create(:principal)
  end

  def then_it_doesnt_blow_up
    expect(admin_firm_page).to_not have_text('NoMethodError')
  end
```

Suggest maybe hiding it for those principals — since it shouldn't occur in normal usage after we've run the rake task but it also probably shouldn't 500

neoeno: (Resolved)

neoeno:
:+1:

rnalexander:
:+1:

PR #279, opened 2015-07-24
scoping firms and trading names to the current user within self-service

JiggyPete:
Original implementation allow signed in self service users to tweak urls to edit other users firms and trading_names.

To solve this we've scoped everything from the `current_user.principal.firm`

### DIFF
--- a/app/controllers/self_service/trading_names_controller.rb
+++ b/app/controllers/self_service/trading_names_controller.rb
@@ -15,6 +15,20 @@ module SelfService
       end
     end
 
+    def edit
+      @firm = principal.firm.trading_names.find(params[:id])
+    end
+
+    def update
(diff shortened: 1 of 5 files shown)


app/controllers/self_service/trading_names_controller.rb
```diff
@@ -15,6 +15,20 @@ def create
       end
     end
 
+    def edit
+      @firm = principal.firm.subsidiaries.find(params[:id])
```
neoeno: we should stick to `principal.firm.trading_names` here — maybe change the usage in `initialize_firm_from_lookup_trading_name` too while we're here

neoeno:
looks good — sad to lose the actions from my `AbstractFirmsController` but it's the only way that really makes sense :+1:

tomas-stefano:
:+1:

PR #280, opened 2015-07-27
Add email to devise permitted params for account update

neoeno:
So that we can change the email address on our user object.

Also remove it from `sign_in` since I added that thinking that it would fix this issue & evidently it did not.


munckymagik:
:+1:

munckymagik:
Re-:+1:

asmega:
:+1:

benbarnett:
image

PR #283, opened 2015-07-29
[DEV APPROVED] Update dough from 5.1.0 -> 5.7.3

neoeno:
RAD pulls in this stuff in a bit of a funny way. Two points.

1. We're calling it from github, not the gem server, so that we don't need VPN access.
2. We also pull in frontend's styles by calling through to a newish version on the CDN.

Questions on this welcome.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'devise_security_extension'
 gem 'dough-ruby',
     github: 'moneyadviceservice/dough',
     require: 'dough',
-    ref: 'cf08913'
+    ref: '77dc86f'
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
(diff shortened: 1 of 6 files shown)


neoeno:
also apparently it doesn't work on travis

benbarnett:
Yep, this was the initial idea. Would love a rake task that did this for you 😊

neoeno:
Me too — though it looks to be a bit tricky to do. We'd have to grab the CDN URLs from either live or QA, and then somehow work out what version of dough should be used alongside it. I'm not sure automating it would pay off time-wise, but I could be wrong.

Any better ideas, @benbarnett @munckymagik?

benbarnett:
Yeah agree, it'd take a long time to recoup the time investment to build something automated. Probably not worth it. Nice things are nice though.

benbarnett:
:+1:

munckymagik:
:+1:

PR #284, opened 2015-07-29
Add a login panel to the RAD homepage

munckymagik:
# Rationale

We will now have many more users visiting this page to sign-in rather than to register. Therefore we need to make it possible to sign-in more easily from the homepage.

# Screenshots

## On desktop:

screen shot 2015-07-29 at 13 52 14

## On mobile:

**Note:** Side column appears above main column on small screens

screen shot 2015-07-29 at 13 53 31

### DIFF
--- a/app/views/principals/pre_qualification_form.html.erb
+++ b/app/views/principals/pre_qualification_form.html.erb
@@ -5,6 +5,10 @@
   </div>
 
   <div class="l-2col">
+    <section class="l-2col-side l-2col-side--pull-right">
+      <%= render 'shared/sign_in_panel' %>
+    </section>
+
     <section class="l-2col-main">
       <%= heading_tag(t('registration.sub_heading'), level: 2) %>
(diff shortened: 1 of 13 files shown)


app/views/principals/pre_qualification_form.html.erb
```diff
@@ -1,119 +1,129 @@
-<div class="l-constrained">
```
munckymagik: This is a terrible diff. Essentially:

* The original content has been **indented** and placed into an `l-2col-main`
* A new `<section>` has been added before it to render the `sign_in_panel` partial
* A new `l-1col` section has been added to place the introductory elements evenly above the columns below.

Any questions, just ask.

benbarnett:
LGTM :+1:

neoeno:
:+1:

munckymagik:
@neoeno can you review the "Fix url matcher ..." commit please?

neoeno:
seems good :+1:

PR #285, opened 2015-07-30
Fix to prevent legend tag content escaping bounds on IE browsers

munckymagik:
Same problem as observed in https://github.com/moneyadviceservice/rad/pull/267 but on a different screen.

Observed on IE8, 9, 10 & 11.

Before:

screen shot 2015-07-30 at 15 16 44

After:

screen shot 2015-07-30 at 15 17 53


benbarnett:
:+1:

Guntrisoft:
:+1:

PR #287, opened 2015-07-31
[WIP] Adding a report to view the uptake on sent invitations

urfolomeus:
Mark Nutley wanted to be able to see what the uptake was like on the invitations that they had sent out to firms. As the uptake looks to have been pretty slow so far, we figured that this report will need to be run reasonably frequently in the near future and so have turned it into a simple rake task.

### DIFF
--- a/app/reports/invitation_uptake.rb
+++ b/app/reports/invitation_uptake.rb
@@ -1,0 +1,35 @@
+class InvitationUptake
+  def run
+    CSV.generate do |csv|
+      csv << ['FCA Number', 'Registered Name', 'Invitation Sent', 'Invitation Accepted', 'Password Reset Sent']
+      query.all.each do |row|
+        csv << [
+          row.fca_number,
+          row.registered_name,
+          nice_date(row.invitation_sent_at),
+          nice_date(row.invitation_accepted_at),
+          nice_date(row.reset_password_sent_at)
+        ]
+      end
+    end
+  end
+
+  private
+
+  def query
+    User.where.not(invitation_sent_at: nil)
+      .joins(principal: [:firm])
+      .select(
+        'firms.fca_number',
+        'firms.registered_name',
+        'users.invitation_sent_at',
+        'users.invitation_accepted_at',
+        'users.reset_password_sent_at'
+      )
+      .order('firms.fca_number')
+  end
+
+  def nice_date(date)
+    date.try(:strftime, '%d-%b-%Y %H:%M')
+  end
+end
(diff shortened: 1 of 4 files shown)


benbarnett:
:+1:

app/reports/invitation_uptake.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def query
+    User.where.not(invitation_sent_at: nil)
+      .joins(principal: [:firm])
+      .select(
+        'firms.fca_number',
+        'firms.registered_name',
+        'users.invitation_sent_at',
+        'users.invitation_accepted_at',
+        'users.reset_password_sent_at'
```
munckymagik: I don't understand what we can assume about the `reset_password_sent_at` field for a user at this point. Maybe worth asking @JiggyPete. 

However, I think there are some other fields we can use to reason about how far they've made it through the invitation process:

1. We could show whether the `password` field is blank or not? I don't believe there is any value set in the password field until they've actually accepted the invite and set one. So this would be a good indicator that they've followed the process to that point.
2. We can find out if they got as far as actually signing in. There are `User.sign_in_count` and `User.last_sign_in_at` fields which would show that they've actually successfully logged in. I think it would be worth putting some test data into the system so we can follow an invite ourselves and see what fields change as we follow the process through.

What I'm expecting to get out should we total up the different types of states here is a funnel for example:

Invitations sent: 10
Invitations accepted: 9
Passwords set: 5
Successful Login least once: 4
Return visits already: 3

urfolomeus: I had a chat with @JiggyPete earlier about the `reset_password_sent_at` and I think it might be a red herring. I concur with your thoughts above on the extra fields. The existing tests are fairly comprehensive so adding more to follow the actual process should be relatively straightforward I think?

Remember as well though that Mark wants to know _who_ the firms are so just including a count is not enough. Prefixing as a TL;DR might be of benefit though if you can find a way to easily add it to the CSV (or I guess you could run a separate report that gives you that).

munckymagik:
Also although we're giving all the raw data so that it can be imported into a spreadsheet, could we also show some kind of summary of the numbers of Users in each stage of the funnel?

urfolomeus:
@munckymagik probably, the question is where? I elected not to put it out through the front end as you're dealing with sensitive info here. Be quicker I'd have though just to either run a separate report that outputs that or just quickly open in Excel and get the counts that way and add to the email you send to Mark.

munckymagik:
@urfolomeus If we output it all in CSV format, with the summary data first then the full table with headers below, would that work?

urfolomeus:
Well it wouldn't be valid CSV but Excel might do the right thing, I'm not sure.

urfolomeus:
Personally I'd leave the CSV as it is and put the highlights in the email. Less chance of confusion.

mconnell:
Closing this PR as it has been superseded by #288, which puts the functionality into the admin UI.

PR #288, opened 2015-08-04
Invitation report in admin area

JiggyPete:
Stakeholders need access to stats on the uptake of inviting firms to maintain their own data via the new self-service area of the site.

This PR makes that data available on the admin area of the site so that the most up to date data can be viewed whenever it is needed.

screen shot 2015-08-04 at 10 40 09


munckymagik:
:+1:

Tr4pSt3R:
:+1:

PR #290, opened 2015-08-05
Get MultiTableFilter responding to input events

neoeno:
E.g. IE11 has a 'clear' button on its input fields. Because this raises `input` rather than `keydown`, the library MultiTableFilter uses to filter ignores it. We'll work around this by mapping input to keydown on our filter fields.

This also fixes pasting into the field.


benbarnett:
:+1:

mconnell:
:key: :arrow_down:

:+1:

PR #291, opened 2015-08-05
Fix adviser ref number field

neoeno:
Initial reported bug was:

> when on the advisers new/edit screen, if the user is entering text in the 'Adviser FCA individual reference number' box when navigating the characters using the arrow keys, the cursor jumps to the right hand end of the text.

This is now fixed with a combination of making the server do the uppercasing, and using CSS `text-transform` to make the text appear uppercase.

I also noticed a few other pieces of code that I didn't write, so I decided to mess them up a bit. Comments explain.


darokel:
:+1:

mconnell:
:+1:

PR #292, opened 2015-08-06
Add ethical and sharia investing fields to firm form

neoeno:
image

It's the last two on the end thar

### DIFF
--- a/spec/support/self_service/firm_edit_page.rb
+++ b/spec/support/self_service/firm_edit_page.rb
@@ -35,6 +35,11 @@ module SelfService
     element :inheritance_tax_and_estate_planning_flag, '.t-inheritance-tax-and-estate-planning-flag'
     element :wills_and_probate_flag, '.t-wills-and-probate-flag'
 
+    element :offers_ethical_investing_true, '.t-ethical-investing-true'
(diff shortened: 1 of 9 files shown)


benbarnett:
:+1:

spec/support/self_service/firm_edit_page.rb
```diff
@@ -47,6 +52,22 @@ def offers_free_initial_meeting?
       !!offers_free_initial_meeting_true.checked?
     end
 
+    def offers_ethical_investing=(choice)
+      send("offers_ethical_investing_#{choice}").set(true)
+    end
+
+    def offers_ethical_investing?
+      !!offers_ethical_investing_true.checked?
```
JiggyPete: I'm wondering why the need for the double bang?
Can `:checked?` return nil?

urfolomeus:
:+1:

spec/support/self_service/firm_edit_page.rb
neoeno: I was following @munckymagik's pattern here — just checked with him and apparently `#checked?` returns something unusual, and we wanted to check against true or false

JiggyPete: makes sense

JiggyPete:
:+1:

PR #293, opened 2015-08-06
Change password page updates

urfolomeus:
Bringing a bit of commonality to the language that we use during the invitation/registration/authentication process.

### DIFF
--- a/app/views/devise/invitations/edit.html.erb
+++ b/app/views/devise/invitations/edit.html.erb
@@ -31,7 +31,7 @@
     <div class="registration__field">
       <%= f.form_row :password_confirmation do %>
         <%= f.errors_for :password_confirmation %>
-        <%= f.label :password_confirmation, t('authentication.common.field_label.password_confirmation'), class: "form__label-heading" %>
+        <%= f.label :password_confirmation, t('authentication.accept_invitation_page.field_labels.password_confirmation'), class: "form__label-heading" %>
         <%= f.password_field :password_confirmation, class: 't-password-confirmation-field', autocomplete: "off", "aria-describedby" => "tooltip_password_confirmation" %>
       <% end %>
     </div>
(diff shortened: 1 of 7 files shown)


app/views/devise/invitations/edit.html.erb
```diff
@@ -31,7 +31,7 @@
     <div class="registration__field">
       <%= f.form_row :password_confirmation do %>
         <%= f.errors_for :password_confirmation %>
-        <%= f.label :password_confirmation, class: "form__label-heading" %>
+        <%= f.label :password_confirmation, t('authentication.accept_invitation_page.field_labels.password_confirmation'), class: "form__label-heading" %>
```
asmega: fyi this should be doing a locale lookup under the hood so you won't need an explicit key. i recommend the i18n-debug gem to see what key look ups are happening.

asmega:
:+1:

app/views/devise/invitations/edit.html.erb
urfolomeus: Thanks, @asmega. Not too sure I follow though. I'm explicitly calling a key here because it is being named differently to the normal password_confirmation on the model (i.e. this is "Confirm new password", not "Confirm password"). Or am I grabbing the wrong end of the stick? :)

urfolomeus: Or are you referring to the fact that I'm not following the standard naming convention for views (which is a very good point and not sure how I missed that).

JiggyPete:
:+1:

app/views/devise/invitations/edit.html.erb
asmega: I'm saying you should not need to add the translation code here. instead by adding the correct key to `config/locales/XX.yml` the correct translation will appear in your view. This is nice to have and is not necessary for this pull request.

urfolomeus: Ah, I see. I was just following the pattern that was already here. If I was to use that method of doing the password confirmation here then I'd be introducing another way of doing the locale lookup, which I was keen not to do in this PR (the rabbit hole was deep enough) :)

urfolomeus: Thanks also @asmega for the heads up on the i18n-debug gem. Good to have in the toolbelt :)

PR #297, opened 2015-08-10
Add required asterisk to initial meeting duration question.

neoeno:
[Sorry for the back-and-forth on this!]

In the past, when we didn't conditionally show and hide this field depending on whether it was applicable, it wouldn't have made sense to mark this as required.

However, now it's only visible when it is required. So we should add an asterisk to it.


benbarnett:
Makes sense :+1:

JiggyPete:
:+1:

PR #298, opened 2015-08-14
[DEV APPROVED] Multiple offices part 1: the index page

munckymagik:
Just the index page functionality for the multiple offices feature we are working on. Below are some screenies of what's been done in this PR.

* First office added is considered the 'Main Office' that will be shown as the primary office on the directory. It is not removable as all firms must have at least one office.
* Add and edit screens will be done in a subsequent PR
* Depends on changes in PR: https://github.com/moneyadviceservice/mas-rad_core/pull/84

## Link from the Firm edit page

screen shot 2015-08-14 at 14 14 38

## Message shown when there are no offices yet

screen shot 2015-08-14 at 14 16 24

## A bunch of office records

screen shot 2015-08-14 at 14 14 07

## Confirmation on hitting the remove button

screen shot 2015-08-14 at 14 15 03

## On-page Filtering

screen shot 2015-08-14 at 14 24 09

### DIFF
--- a/app/assets/stylesheets/components/_offices_table.scss
+++ b/app/assets/stylesheets/components/_offices_table.scss
@@ -1,27 +1,27 @@
 .offices-table {
   width: 100%;
+}
 
-  .offices-table__address {
-    width: auto;
-  }
+.offices-table__address {
+  width: auto;
+}
 
-  .offices-table__address-postcode {
-    width: 8%;
-  }
+.offices-table__address-postcode {
+  width: 8%;
+}
 
-  .offices-table__telephone-number {
-    width: 12%;
-  }
+.offices-table__telephone-number {
+  width: 12%;
+}
 
-  .offices-table__email-address {
-    width: auto;
-  }
+.offices-table__email-address {
+  width: auto;
+}
 
-  .offices-table__disabled-access {
-    width: 8%;
-  }
+.offices-table__disabled-access {
+  width: 8%;
+}
 
-  .offices-table__action {
-    width: 12%;
-  }
+.offices-table__action {
+  width: 12%;
 }
--- a/app/assets/stylesheets/components/_offices_table.scss
+++ b/app/assets/stylesheets/components/_offices_table.scss
@@ -1,0 +1,27 @@
+.offices-table {
+  width: 100%;
+
+  .offices-table__address {
+    width: auto;
+  }
+
+  .offices-table__address-postcode {
+    width: 8%;
+  }
+
+  .offices-table__telephone-number {
+    width: 12%;
+  }
+
+  .offices-table__email-address {
+    width: auto;
+  }
+
+  .offices-table__disabled-access {
+    width: 8%;
+  }
+
+  .offices-table__action {
+    width: 12%;
+  }
+}
--- a/app/controllers/self_service/offices_controller.rb
+++ b/app/controllers/self_service/offices_controller.rb
@@ -4,6 +4,9 @@ module SelfService
     before_action -> { @firm = current_firm }
     before_action -> { @office = current_office }, only: [:destroy]
 
+    def index
+    end
+
     def destroy
       @office.destroy
       flash[:notice] = I18n.t('self_service.office_destroy.deleted', postcode: @office.address_postcode)
--- a/app/controllers/self_service/offices_controller.rb
+++ b/app/controllers/self_service/offices_controller.rb
@@ -1,0 +1,5 @@
+module SelfService
+  class OfficesController < ApplicationController
+    before_action :authenticate_user!
+  end
+end
--- a/app/controllers/self_service/offices_controller.rb
+++ b/app/controllers/self_service/offices_controller.rb
@@ -2,6 +2,14 @@ module SelfService
   class OfficesController < ApplicationController
     before_action :authenticate_user!
     before_action -> { @firm = current_firm }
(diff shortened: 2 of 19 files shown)


app/assets/stylesheets/components/_offices_table.scss
```diff
@@ -0,0 +1,27 @@
+.offices-table {
+  width: 100%;
+
+  .offices-table__address {
```
urfolomeus: Should these be nested? I thought the BEM way was not use naming (as you have) rather than nesting?

munckymagik: Well spotted. The same thing has been done in _advisers_table.scss, I may fix there too. Asking the authors.

benbarnett: :+1:

benbarnett:
:+1:

app/views/self_service/offices/_offices_table.html.erb
```diff
@@ -0,0 +1,48 @@
+<div class="table-wrapper">
+  <table class="offices-table">
```
urfolomeus: Should this also use `.datatable-default` to make it a scrollable table or is it unlikely to contain that many values?

urfolomeus: Sorry, just noticed the filter. Ignore that comment. :)

app/helpers/self_service/self_service_helper.rb
```diff
@@ -20,9 +20,29 @@ def add_adviser_button(firm:)
...
+    def office_address_table_cell(office)
+      [
+        :address_line_one,
+        :address_line_two,
+        :address_town,
+        :address_county
+      ].map { |field| office.send(field) }
+        .reject(&:blank?)
+        .join(', ')
+    end
```
urfolomeus: This is lovely. <3 :)

app/views/self_service/offices/_offices_table.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <tr class="t-office-table-row">
+          <td class="t-address" data-dough-filterable><%= office_address_table_cell(office) %></td>
+          <td class="t-address-postcode" data-dough-filterable><%= office.address_postcode %></td>
+          <td class="t-telephone-number" data-dough-filterable><%= office.telephone_number %></td>
+          <td class="t-email-address" data-dough-filterable><%= office.email_address %></td>
+          <td class="t-disabled-access"><%= office.disabled_access ? 'Yes' : 'No' %></td>
+          <td>
+            <% if office == @firm.main_office %>
+              <span class="t-main-office">Main Office</span>
+            <% else %>
```
urfolomeus: How do you delete a main office? It doesn't look like you can make another office the main one, so does that not make it impossible to delete this one? Or will future functionality handle this?

urfolomeus:
Some minor points but otherwise :+1: :)

app/helpers/self_service/self_service_helper.rb
munckymagik: Oh good :)

app/controllers/self_service/offices_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    private
+
+    def principal
+      current_user.principal
+    end
+
+    def current_firm
+      Firm.find_by(id: params[:firm_id], fca_number: principal.fca_number)
```
JiggyPete: If firm 123 is within the self-service area, they could access their offices then tweak the :firm_id and be able to access another firms offices.

`current_firm` should be scoped to `principal`

app/controllers/self_service/offices_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    before_action -> { @firm = current_firm }
+    before_action -> { @office = current_office }, only: [:destroy]
+
+    def destroy
+      @office.destroy
+      flash[:notice] = I18n.t('self_service.office_destroy.deleted', postcode: @office.address_postcode)
+
+      redirect_to :back
+    end
+
```
JiggyPete: There is an implicit `:index` action. For clarity I'd suggest adding it in.

app/views/self_service/offices/_offices_table.html.erb
munckymagik: You don't delete it. There must always be one office. If the details need to change then the user can edit them.

app/controllers/self_service/offices_controller.rb
munckymagik: The `find_by` is also querying based on `principal.fca_number`. Does that not provide the scoping we need or have I missed something?

app/views/self_service/offices/_offices_table.html.erb
munckymagik: And also, because this is a step on the way, it may not be obvious that the address etc. in the Firm records is going to be removed and migrated to become this first office record.

app/controllers/self_service/offices_controller.rb
JiggyPete: D'oh! You are spot on. Anytime I see `Firm.find` I get worried.

app/controllers/self_service/offices_controller.rb
munckymagik: Ok yes. Added in 683cf79

JiggyPete:
:+1:

PR #300, opened 2015-08-17
Fix advisers_edit_spec, it was using the wrong variable

munckymagik:
The test fixture was being created using the data we expected to change it to in the test that checks that saving changes actually works.


neoeno:
:+1:

benbarnett:
:+1:

PR #303, opened 2015-08-20
[DEV APPROVED] Language Selector

neoeno:
Firms want to be able to display languages they can give advice in. Here's a component that lets them do that.

It'll probably still be subject to some iteration on UX and other feedback, but I think this is good to merge.

## How it is

Example

## The designs

Designs

### DIFF
--- a/app/assets/javascripts/modules/LanguageSelector.js
+++ b/app/assets/javascripts/modules/LanguageSelector.js
@@ -1,0 +1,103 @@
+/**
+ * # Language selector
+ *
+ * A set of select boxes that let the user select multiple languages.
+ *
+ * See test fixture (/spec/javascripts/fixtures/LanguageSelector.html)
+ * for example markup.
+ *
+ * @module LanguageSelector
+ * @returns {class} LanguageSelector
+ */
+
+define(['jquery', 'DoughBaseComponent'],
+       function($, DoughBaseComponent) {
+  'use strict';
+
+  var LanguageSelectorProto,
+      defaultConfig = {
+        messageAttribute: 'data-dough-confirmation-message',
+        selectorPrefix: 'data-dough-language-selector'
+      };
+
+  /**
+   * @constructor
+   * @extends {DoughBaseComponent}
+   * @param {HTMLElement} $el    Element with dough-component on it
+   * @param {Object}      config Hash of config options
+   */
+  function LanguageSelector($el, config) {
+    LanguageSelector.baseConstructor.call(this, $el, config, defaultConfig);
+  }
+
+  /**
+   * Inherit from base module, for shared methods and interface
+   */
+  DoughBaseComponent.extend(LanguageSelector);
+  LanguageSelector.componentName = 'LanguageSelector';
+  LanguageSelectorProto = LanguageSelector.prototype;
+
+  /**
+   * Init function
+   *
+   * Set up listeners and accept promise
+   *
+   * @param {Object} initialised Promise passed from eventsWithPromises (RSVP Promise).
+   * @returns {LanguageSelector}
+   */
+  LanguageSelectorProto.init = function(initialised) {
+    this.bindToElements();
+    this._initialisedSuccess(initialised);
+    return this;
+  };
+
+  /**
+   * Bind to events on elements
+   *
+   * Just the add and delete language links.
+   */
+  LanguageSelectorProto.bindToElements = function() {
+    this.$find('add-language').on('click', $.proxy(this.addLanguage, this));
+    this.$el.on('click', this._buildSelectorQuery('delete-language'), $.proxy(this.deleteLanguage, this));
+  };
+
+  /**
+   * Add a new language selector to the list of selectors
+   *
+   * Employs a template and clears all relevant attributes from it
+   * before cloning and inserting.
+   */
+  LanguageSelectorProto.addLanguage = function() {
+    var $container = this.$find('selectors').first(),
+        $template = this.$find('template').clone();
+    $template.find('select')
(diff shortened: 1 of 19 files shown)


app/assets/stylesheets/components/_language_selector.scss
```diff
@@ -0,0 +1,16 @@
+.language-selector__select {
+  margin-bottom: $baseline-unit*2;
+}
+
+.language-selector__delete {
+  margin-left: 0.5em;
+  cursor: pointer;
```
neoeno: Should these really be link tags? The designs make them look like links, but maybe they shouldn't actually be links? Since they don't technically hyperlink anywhere. Or is that OK?

Advice appreciated.

benbarnett: Maybe a `button`, even if they look like links? Although it doesn't take you somewhere, it does "do" something?

config/locales/self_service/firm_form.en.yml
```diff
@@ -58,3 +58,8 @@ en:
 
       investment_sizes_heading: Total pension pot / investment size
       investment_sizes_description: Please tell us what size pension pot(s) you are willing to provide advice on. Check all that apply.
+
+      languages_heading: Which langues (other than English) can your firm's advisers offer advice in?
```
neoeno: Langues. Ha

app/assets/javascripts/modules/LanguageSelector.js
```diff
@@ -0,0 +1,103 @@
...
+      };
+
+  /**
+   * @constructor
+   * @extends {DoughBaseComponent}
+   * @param {HTMLElement} $el    Element with dough-component on it
+   * @param {Object}      config Hash of config options
+   */
+  function LanguageSelector($el, config) {
+    LanguageSelector.baseConstructor.call(this, $el, config, defaultConfig);
```
benbarnett: boo ES5

app/assets/stylesheets/components/_language_selector.scss
neoeno: Don't suppose we have a handy button-as-link style hanging around anywhere?

benbarnett: Been digging around, I'm sure we had one before! Will keep looking...

benbarnett: Check this out btw, see if you come to the same conclusion? https://css-tricks.com/use-button-element/

neoeno: I'd tend to agree. The only fly in the ointment is that it's within a form, and buttons in a form, by default, submit that form — so we'll have to interfere with that behaviour. 

Before I've also encountered the first button in a form receiving a click event when the form is submitted by hitting enter on a text field elsewhere — but that was in pensions calculator so could have been for many other reasons...

Might give it a go anyway

neoeno: Solved my own problem I think, `<button type="button">` avoids that behaviour

munckymagik:
:+1: like it.

app/assets/stylesheets/components/_language_selector.scss
neoeno: Have made the change. There was a class — `button-link`. Not sure of its providence exactly.

spec/javascripts/tests/LanguageSelector_spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+    it('clears attributes from select element', function() {
+      expect(this.$selectorsContainer.find('select').last().attr('id')).to.eq('');
+    });
+
+    it('does not duplicate template', function() {
+      expect(this.$html.find('[data-dough-language-selector-template]').length).to.eq(1);
+    });
+
+    context('when "remove language" is clicked', function() {
```
benbarnett: Are we missing a bracket or indentation out or something? The remove language context seems like it's within the add language one.

neoeno: It is! We add one so we can remove one

app/views/self_service/firms/questionnaire/_languages.html.erb
```diff
@@ -0,0 +1,32 @@
...
+
+      <% f.object.languages.each do |language| %>
+        <div class="language-selector__select" data-dough-language-selector-selector>
+          <%= firm_language_select language, class: 't-languages' %>
+          <%= firm_language_delete_button %>
+        </div>
+      <% end %>
+    </div>
+
+    <button type="button" class="button-link language-selector__add" data-dough-language-selector-add-language>
```
benbarnett: Cool. Strictly progressively enhancementy speaking we'd add these elements with javascript as they don't do anything without it. Can probably get away with this as it's the admin interface though.

spec/javascripts/tests/LanguageSelector_spec.js
benbarnett: Ah ha!

benbarnett:
:+1:

app/views/self_service/firms/questionnaire/_languages.html.erb
neoeno: Good point — I've added some CSS and such to make the whole thing behave better without JS. It just shows a single select box now, no non-functional buttons.

neoeno:
@benbarnett That latest commit still good by you?

benbarnett:
Yep looks good!

neoeno:
Thanks! All pending this one now: https://github.com/moneyadviceservice/mas-rad_core/pull/86

PR #304, opened 2015-08-20
[DEV APPROVED] Multiple offices UI (part deux) - the add and edit pages

munckymagik:
screen shot 2015-08-20 at 15 25 47

screencapture-localhost-3000-self_service-firms-153-offices-1-edit-1440080632573

screencapture-localhost-3000-self_service-firms-153-offices-1440080560344

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
-gem 'mas-rad_core', '0.0.70'
+gem 'mas-rad_core', '0.0.71'
 gem 'oga'
 gem 'pg'
 gem 'ransack'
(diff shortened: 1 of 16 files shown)


benbarnett:
:+1:

urfolomeus:
First comment without reviewing code ... is there a way to swap which office is the main one or is that an upcoming feature?

munckymagik:
No. It's fixed as the first one. If you want to swap which one is the main one you can edit the record. This small compromise makes the logic/ui/interactions vastly simpler in several ways.

munckymagik:
... and we've approved this decision with the design/product folks. So all good.

urfolomeus:
Cool, just thought I'd check as it's impossible to delete the main firm.

asmega:
:+1:

urfolomeus:
:+1:

jonalport:
Is it necessary to say that a field is invalid if you've already said it can't be blank?

munckymagik:
@jonalport very good point. Repetitive isn't it. That can be addressed in mas-rad_core. I think a lot of the existing models will/may behave the same so it would have to be a sweeping change not directly related this piece of work. So I'll raise a housekeeping task so someone has a look at it.

PR #305, opened 2015-08-21
Fixing layout

urfolomeus:
As we were working on the notifications story we realised that the base layout was not following a common styling. For example, we are using the `l-registration` class on the devise views in order to match front end styles. However our own content wrappers such as `l-self-service` and `l-pre-qualification` were using different styling.

We were also in some cases nesting `l-constrained` because it wasn't obvious which views were using them and which were not. As such we've hoisted the `l-constrained` to the layout to make it more obvious.

Finally we had to use a named yield for the breadcrumbs because they were now being rendered inside the `l-constrained` due to where they were being called.

**N.B. the views look like there are a lot of changes due to us changing the indentation when we either extracted the `l-constrained` div wrapper (in the case of the principal views) or added the `l-registration` div wrapper (in the case of the devise views)**

### DIFF
--- a/app/assets/stylesheets/_enhanced.scss
+++ b/app/assets/stylesheets/_enhanced.scss
@@ -9,6 +9,7 @@
 
 @import 'layouts/2col';
 @import 'layouts/constrained';
+@import 'layouts/content';
 @import 'layouts/footer';
 @import 'layouts/half_width';
 @import 'layouts/header';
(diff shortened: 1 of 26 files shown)


neoeno:
This changes the way the onboarding notices are displayed from full-width to constrained. If this was your intention that's fine though.

If you're going to use a `content_for` for breadcrumbs, I'd suggest adding `s/<%= render_breadcrumbs/<% render_breadcrumbs` to this, or in some other way make clear that the breadcrumbs aren't part of the template anymore.

:+1:

urfolomeus:
Yes the intention is to have both the flash messages and onboarding messages constrained rather than full-width. This was from the designs sent over by Phil, I believe.

Good point on the `<%=` vs `<%` though. I can never remember which way round they go and I had about 3 different solutions before landing on this one :D

urfolomeus:
@neoeno done :)

neoeno:
:+1:

benbarnett:
:+1:

urfolomeus:
Thanks @benbarnett :)

PR #308, opened 2015-09-01
Fix validations

neoeno:
See commits.

@paulanthonywilson could you QA this (Principal registration flow)?


asmega:
:+1:

paulanthonywilson:
:+1: Validations on Principal are now correct (apart from password validation grammar).

PR #309, opened 2015-09-03
Introducing notifications and new flash messages

JiggyPete:
In this PR we have introduced the onboarding notifications, which appear within the self_service area until the user/firm has registered their firm, added an adviser and an office.

**PLEASE NOTE**

The design called for a common notifications area that included both the flash messages and any other notifications (of which the onboarding notifications are one type). Currently there is an issue where the layout in this project has a nested `l-constrained` (i.e. there is a `l-constrained` in the layout and several top level divs in the views also extend `l-constrained`) and so the `l-flashes` div has mimicked this just now. However, this should probably be rectified to have a single `l-constrained`. We have shied away from making that change in this PR as there is already enough going on. It needs to be decided whether to fix that issue first on a separate branch and then rebase into this one, or whether we go with things as they are just now and then fix it all at once later.

screen shot 2015-09-03 at 11 44 16

We also updated the flash messages to be inline with this new design.

screen shot 2015-09-03 at 11 45 44

screen shot 2015-09-03 at 11 46 57


app/assets/stylesheets/_enhanced.scss
```diff
@@ -34,7 +36,8 @@
 @import 'components/navigation';
 @import 'components/notice/all';
 @import 'components/offices_table';
-@import 'components/onboarding_message';
+@import 'components/form_section';
```
mconnell: duplication of lines 33 + 34

app/assets/stylesheets/layouts/_notifications.scss
```diff
@@ -0,0 +1,54 @@
+$notification-radius: 5px;
+
+.global-alert {
+  border-radius: $notification-radius;
+  margin-top: 1.5em;
```
mconnell: Think the preference is to use `$baseline-unit` (or a multiple of) instead of `em` values

Gemfile
```diff
@@ -24,7 +24,7 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
-gem 'mas-rad_core', '0.0.75'
+gem 'mas-rad_core', '0.0.76'
```
mconnell: I'm assuming the bump is to get access to new rad_core functionality required for this to work?

mconnell:
Other than the comments, it looks nice bit of functionality :+1:

app/assets/stylesheets/layouts/_notifications.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.global-alert__content-container--success:before {
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-image: image-url('icon-tick.svg');
+}
+
+.global-alert__content-container--onboarding:before {
+  background-color: #F89E54;
```
spencer: Probably should be a named colour.

Gemfile
JiggyPete: Yip, there is a new `:onboarded?` method on principal that we use to determine whether to display the onboarding message.

app/assets/stylesheets/layouts/_notifications.scss
JiggyPete: Where would the named colour live?

app/assets/stylesheets/_enhanced.scss
urfolomeus: Good catch!

app/assets/stylesheets/layouts/_notifications.scss
urfolomeus: Yup, will change.

app/assets/stylesheets/layouts/_notifications.scss
urfolomeus: Inside the _variables.scss file :)

JiggyPete:
All updated @mconnell + @spencer.

mconnell:
:+1:

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1 @@
+$warning_orange: #F89E54;
```
neoeno: In general components should own their own variables, and should be specified with hyphens (rather than underscores). [There's an example of this pattern in use here.](https://github.com/moneyadviceservice/rad/tree/master/app/assets/stylesheets/components/notice) And a little explanation [here](https://github.com/moneyadviceservice/rad/pull/198#discussion-diff-31034253) and [here](https://github.com/moneyadviceservice/rad/pull/198#discussion-diff-31034281).

app/assets/stylesheets/layouts/_notifications.scss
```diff
@@ -0,0 +1,54 @@
+$notification-radius: 5px;
```
neoeno: There are a few things in this file.

* See prior comment about where sass variables should live.
* This is a BEM component (no `.l-` prefix) and so belongs in `components` rather than `layouts`
* The file should be named according to the component's name, so `_global_alert.scss`
* It might be an idea to rename it entirely, to something local to the project. It's not really an extension of `global-alert` elsewhere in MAS, so it should probably have a different name to disambiguate.
* There are a few long selectors in this file. BEM ought to mean you only ever have to specify one class (in theory) — so it seems like you're maybe fighting specificity here? If so, can that be resolved in some other way? Renaming the component might help if it's frontend that's interfering.

There's a bit of repetition in here that could probably be extracted too but it's no big deal.

app/helpers/self_service/self_service_helper.rb
```diff
@@ -45,9 +46,12 @@ def office_address_table_cell(office)
...
 
-    def render_onboarding_message(page)
-      render partial: "self_service/onboarding/#{page}",
-             locals: { principal: current_user.principal }
+    def render_onboarding_message
+      return if current_user.principal.onboarded?
+
+      content_for :onboarding_message do
+        render 'shared/onboarding_message'
+      end
```
neoeno: Something seems a little indirect about this. Is there any significant difference in outcome if we ditch the `content_for` and render it directly?

app/views/shared/_onboarding_message.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="global-alert global-alert--onboarding">
+  <div class="global-alert__content-container global-alert__content-container--onboarding">
+    <strong class="global-alert__heading"><%= t('self_service.onboarding.title') %></strong>
+      <%= content_tag :p, t('self_service.onboarding.message').html_safe, class: 'global-alert__message' %>
```
neoeno: Probably don't need the `html_safe` here

app/views/shared/_alerts.html.erb
```diff
@@ -1,6 +1,6 @@
-<% flash.to_hash.slice('alert', 'info', 'notice', 'error').each do |name, msg| %>
+<% flash.to_hash.slice('notice', 'success', 'info', 'alert', 'error').each do |name, msg| %>
   <div class="global-alert global-alert--<%= name %>">
-    <div class="global-alert__content-container">
+    <div class="global-alert__content-container global-alert__content-container--<%= name %>">
       <strong class="global-alert__heading"><%= name.titleize %></strong>
       <%= content_tag :p, msg.html_safe, class: 'global-alert__message t-flash-message' %>
```
neoeno: Actually, is there any good reason for having the `html_safe` here either?

config/locales/self_service.en.yml
```diff
@@ -43,3 +43,7 @@ en:
       breadcrumb: "%{firm_name}"
       title: "Edit: %{firm_name}"
       saved: Saved successfully
+
+    onboarding:
+      title: Onboarding
```
neoeno: I guess this is pending copy? Might be worth mentioning

neoeno:
I like these flash messages a lot more. Maybe they'll get extracted and reused elsewhere! I hope so.

A lot of comments I know — & sorry I got to this late after you'd already addressed first round of review. We can go over it in a call tomorrow (Friday) if you'd like?

app/helpers/self_service/self_service_helper.rb
urfolomeus: @neoeno this is because we want to render the onboarding notifications within the same area as the rest of the notifications (i.e. flashes).

config/locales/self_service.en.yml
urfolomeus: I'd be inclined to let the customer tell us whether or not to change this during the demo. It will draw attention to itself if it needs changed.

app/views/shared/_alerts.html.erb
urfolomeus: I don't know. I would have thought that all content in here would be internal to the app, but maybe not.

app/helpers/self_service/self_service_helper.rb
urfolomeus: For clarity, if we move it inline we get extra spacing between the notifications. Although we could remove that with CSS, my concern is that we are then separating those notifications conceptually.

### Using `content_for`

screenshot 2015-09-04 15 00 09

### Using directly

screenshot 2015-09-04 15 01 27

darokel:
:+1:. This should probably get a pair of frontend eyes too cc @jonalport @nmiddleweek

nmiddleweek:
Seems to be good with me... I don't know RAD so I can't comment on what else may be impacted.

Only question that jumps to mind is the use of .svg for background-image. I know @jonalport has adopted an inline SVG sprites approach and referenced each icon with the USE tag. It resembles an embedded SVG sprite approach with no extra HTTP requests.

I'm wondering if you're approach is more IE friendly?

JiggyPete:
We started with .png in the css, then introduced SVG and USE elements with the help of a spritesheet. 

However it was easier and tidier to separate the svg icons out and use them directly within the css **(especially as they were already using :before for this)**

Re: IE. We've checked in IE 9, 10 and 11 and all looks good with the current background-image implementation. 

IE 8 doesn't support the background image approach or inline html approach. The messages still look good with the red/green/orange background all the same.

app/assets/images/icon-cross.svg
```diff
@@ -0,0 +1,12 @@
...
+    <!-- Generator: Sketch 3.3.3 (12072) - http://www.bohemiancoding.com/sketch -->
+    <title>icon-cross</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
+        <g id="icon-cross" sketch:type="MSArtboardGroup" fill="#FFFFFF">
+            <path d="M10.001768,5 L5.00176803,0 L0,5 L5.00176803,10 L0,15 L5.00176803,20 L10.001768,15 L15.001768,20 L20.001768,15 L15.001768,10 L20.001768,5 L15.001768,0 L10.001768,5 Z" id="Imported-Layers-Copy" sketch:type="MSShapeGroup"></path>
+        </g>
+    </g>
+</svg>
```
benbarnett: All these SVGs could do with optimising – could you run them through https://jakearchibald.github.io/svgomg/?

app/assets/stylesheets/components/rad-notification/_rad-notification.scss
```diff
@@ -0,0 +1,66 @@
...
+    background-position: center center;
+    background-image: image-url('icon-tick.svg');
+
+    @include respond-to($mq-m) {
+      width: $baseline-unit*8;
+    };
+  }
+}
+
+.rad-notification__content-container--error:before, .rad-notification__content-container--alert:before {
```
benbarnett: Can we drop each selector onto its own line?

app/assets/stylesheets/components/rad-notification/_settings.scss
```diff
@@ -0,0 +1,5 @@
+$rad-notification-radius: 5px;
+$color-rad-notification-warning: #F89E54;
```
benbarnett: Could this reference another variable in the main palette for RAD? Could potentially be reused?

app/views/layouts/application.html.erb
```diff
@@ -28,11 +28,14 @@
 <body>
 <%= render 'shared/skip_links' %>
 <%= render 'shared/header' %>
-<%= render 'shared/alerts' %>
 <%= yield :breadcrumbs %>
 
 <main role="main" id="main">
   <div class='l-constrained'>
+    <div class="l-flashes">
```
benbarnett: Can we move this div into the partial, or make it a conditional somehow? Just thinking that we were setting the tone for how to do flash messages across MAS and having an empty div when there are no flash messages seems like it could be improved

app/views/shared/_alerts.html.erb
```diff
@@ -1,8 +1,8 @@
-<% flash.to_hash.slice('alert', 'info', 'notice', 'error').each do |name, msg| %>
-  <div class="global-alert global-alert--<%= name %>">
-    <div class="global-alert__content-container">
-      <strong class="global-alert__heading"><%= name.titleize %></strong>
-      <%= content_tag :p, msg.html_safe, class: 'global-alert__message t-flash-message' %>
+<% flash.to_hash.slice('notice', 'success', 'info', 'alert', 'error').each do |name, msg| %>
+  <div class="rad-notification rad-notification--<%= name %>">
```
benbarnett: Can the notification markup be moved into a helper or something? Feels like there's a bit of duplication going on.

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1 @@
+
```
benbarnett: Is this needed? I suppose you could move that colour variable mentioned in another comment into here?

app/assets/stylesheets/components/rad-notification/_settings.scss
JiggyPete: We've moved this to the main _settings file in RAD. Is it ok to leave this here until there is a need for it elsewhere, or would you prefer it to be added to Dough just now?

benbarnett: Definitely fine as is but I think there's a slight misunderstanding.

We should still have the 'local' definition of this variable in the component's `_settings.scss` file, but it should be referencing a global colour that lives in the project's `variables.scss`. Does that make sense?

app/views/layouts/application.html.erb
JiggyPete: We alluded to this in the main PR message above (although granted it's not clear). Basically we're having to wrap the flashes and notifications in a div just now for alignment with the logo and the breadcrumbs, which are getting their styles from frontend. This div is a temporary splint until we deal with the root cause. Ideally we don't want to make that whole fix in this PR, as it is already pretty big.

It's a bit complicated to explain what all we think needs to be done in a comment, but perhaps we can demo this to you at the frontend meeting today?

benbarnett: Sure, thought that might be the case. A demo could be good in the meeting today actually!

app/assets/stylesheets/components/rad-notification/_settings.scss
JiggyPete: Gotcha

JiggyPete:
@benbarnett can you take a look at the new commits please?

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1 @@
+$color-orange-light: #F89E54;
```
benbarnett: Just spotted we now have a variables and a settings file doing the same thing. We should merge them into `variables.scss`

app/assets/stylesheets/components/rad-notification/_settings.scss
```diff
@@ -0,0 +1,5 @@
+$rad-notification-radius: 5px;
+$color-rad-notification-warning: $color-orange-light;
+$color-rad-notification-message-bg: $color-global-alert-message-bg;
+$color-rad-notification-default: $color-global-alert-default;
+$color-rad-notification-error: $color-global-alert-error;
```
benbarnett: Better :+1:

benbarnett:
Looks good after merging variables and settings :+1:

urfolomeus:
@benbarnett thanks _settings.scss -> _variables.scss

app/assets/stylesheets/_settings.scss
```diff
@@ -1 +0,0 @@
-$background-color--light: $color-grey-paler;
```
benbarnett: Can't tell from this diff but is this file actually deleted?

urfolomeus: Yeah, if you go to files changed view and then click the "34 changed files" link you can see it's been removed.

benbarnett: Sounds good :+1:

benbarnett:
:+1:

urfolomeus:
Thanks @benbarnett :)

PR #312, opened 2015-09-09
By country report

urfolomeus:
This was a (so far) one-off report to show the uptake of the register by country.
It's a little rough but has been committed in case we need it again.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,6 +32,7 @@ AllCops:
   - 'bin/**/*'
   - 'config/**/*'
   - 'db/schema.rb'
+  - 'lib/reports/by_country_report.rb'
   - 'node_modules/**/*'
   - 'tmp/**/*'
   - 'vendor/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,7 +32,6 @@ AllCops:
   - 'bin/**/*'
   - 'config/**/*'
   - 'db/schema.rb'
-  - 'lib/reports/by_country_report.rb'
   - 'node_modules/**/*'
   - 'tmp/**/*'
   - 'vendor/**/*'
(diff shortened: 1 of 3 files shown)


JiggyPete:
:+1: Good to keep this incase we need to do it again

lib/reports/by_country_report.rb
```diff
@@ -0,0 +1,94 @@
...
+      @lookup[postcode_slug_for(adviser.firm)]
+    end
+    generate_report('Advisers', grouped_advisers, advisers.count)
+  end
+
+  def error_report
+    grouped_firms = @registered_firms.group_by do |firm|
+      @lookup[postcode_slug_for(firm)]
+    end
+    firms_with_errors = grouped_firms.select { |k, _| k.blank? }.values.flatten.map(&:id).uniq
```
asmega: why do these firms have errors? i.e. why would `k` be blank?

urfolomeus: `k` will be blank for any postcodes which return a blank country (i.e. those that are not found on the postcode lookup).

.rubocop.yml
```diff
@@ -32,6 +32,7 @@ AllCops:
   - 'bin/**/*'
   - 'config/**/*'
   - 'db/schema.rb'
+  - 'lib/reports/by_country_report.rb'
```
munckymagik: Why doesn't this file pass?

urfolomeus: Length of the `fetch_countries` method. Peter and I tried to get it down below 15 lines yesterday and managed to get 1 line off by making it unreadable. :)

Decided it wasn't worth sinking more time into it and ignored the file instead.

urfolomeus: Actually I'm going to have another go whilst I finish my lunch.

urfolomeus: OK @munckymagik, this should be (a bit) better now.

munckymagik:
:+1:

asmega:
:+1:

PR #313, opened 2015-09-10
Removing l-flashes and wrapping divs everywhere

JiggyPete:
This PR is to allow easier lifting of the new flash message style from RAD into Dough by removing the wrapping l-flashes div within application.html.erb.

There should be no noticeable change to the user.

This involved removing the 
* l-pre-qualification class and wrapping div from the prequalification page
* l-registration class and wrapping div from the devise views
* l-self-service class and wrapping div from the self service views

There is also a helper which switches the top level class between l-content and l-registration.

At a glance, some of the diffs look a larger than they are. In the vast majority of html.erb files we've just removed a wrapping div and unindented.

### DIFF
--- a/app/assets/stylesheets/layouts/_content.scss
+++ b/app/assets/stylesheets/layouts/_content.scss
@@ -1,6 +1,6 @@
 // This mimics the Front End l-registration style as we needed to use that for our user registration
 // https://github.com/moneyadviceservice/frontend/blob/master/app/assets/stylesheets/layout/page_specific/_registration.scss#L1
-%l-content {
+.l-content {
   @include column(12);
   margin-bottom: $baseline-unit*4;
 }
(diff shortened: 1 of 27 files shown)


app/assets/stylesheets/layouts/_content.scss
```diff
@@ -4,3 +4,7 @@
   @include column(12);
   margin-bottom: $baseline-unit*4;
 }
+
+.l-content {
+  @extend %l-content;
+}
```
urfolomeus: Now that we're not using the abstract class we should replace it with this concrete one.

urfolomeus:
Apart from that one wee comment :+1: :)

app/assets/stylesheets/layouts/_content.scss
JiggyPete: Well spotted, thanks!

benbarnett:
As you say the diff looks slightly more major than it actually is, but I can't see anything untoward. Looks good :+1:

JiggyPete:
Thanks @urfolomeus + @benbarnett!

PR #317, opened 2015-09-30
[DEV APPROVED] Add a rake task to generate some stats requested by the CEO

munckymagik:
Requested today. Keeping this around in case it makes things quicker in future.

### DIFF
--- a/lib/tasks/adhoc.rake
+++ b/lib/tasks/adhoc.rake
@@ -11,4 +11,32 @@ namespace 'adhoc' do
     puts report.adviser_report
     puts report.error_report
   end
+
+  desc 'Prints several interesting stats about the firms in the directory'
+  task firm_stats: :environment do
+    published_firms = Firm.registered.select(&:publishable?)
+
+    puts "Total firms (face-to-face & remote): #{published_firms.count}"
+    remote = published_firms.select { |f| f.primary_advice_method == :remote }
+    puts "Total firms (remote): #{remote.count}"
+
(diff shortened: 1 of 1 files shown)


lib/tasks/adhoc.rake
```diff
@@ -11,4 +11,32 @@ namespace 'adhoc' do
...
+    remote = published_firms.select { |f| f.primary_advice_method == :remote }
+    puts "Total firms (remote): #{remote.count}"
+
+    no_min_fee = published_firms.select { |f| [0, nil].include?(f.minimum_fixed_fee) }
+    puts "Firms with NO minimum fee: #{no_min_fee.count}"
+
+    fee_between_1_and_500 = published_firms.select { |f| (1..500).include?(f.minimum_fixed_fee) }
+    puts "Firms who have a minimum fee between £1.00 - £500: #{fee_between_1_and_500.count}"
+
+    fee_between_501_and_1000 = published_firms.select { |f| (501..1000).include?(f.minimum_fixed_fee) }
```
asmega: would this be better if the `select` was a `where` instead. so the database does the lifting instead of ruby?

munckymagik: Good point. However, the report runs in an acceptable amount of time (just a few seconds) so in this case I've chosen not to optimise. `¯\_(ツ)_/¯`

asmega: :+1:

pmcostello: :+1:

PR #319, opened 2015-10-06
Change flash messages

munckymagik:
Updates the bold headings for the flash messages:
pasted image at 2015_10_06 11_27 am

... and also updates the individual flash messages for added/saved records.

### DIFF
--- a/spec/controllers/selfservice/advisers_controller/create_spec.rb
+++ b/spec/controllers/selfservice/advisers_controller/create_spec.rb
@@ -23,7 +23,7 @@ module SelfService
         end
 
         it 'adds a success flash message' do
-          expect(flash[:notice]).to eq('Saved successfully')
+          expect(flash[:notice]).to eq(I18n.t('self_service.adviser_edit.saved'))
         end
       end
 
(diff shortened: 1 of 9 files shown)


Tr4pSt3R:
Awesome :+1:

spec/controllers/selfservice/advisers_controller/create_spec.rb
```diff
@@ -23,7 +23,7 @@ module SelfService
         end
 
         it 'adds a success flash message' do
-          expect(flash[:notice]).to eq('Saved successfully')
+          expect(flash[:notice]).to eq(I18n.t('self_service.adviser_edit.saved'))
```
rnalexander: Total stickler point here.  Any reason why you're using the 'I18n.t' notation here but the shorter 't' notation in the _alerts.erb file?

munckymagik: IIRC the shorter `t` version is a helper in the views and thus doesn't need a prefix. Elsewhere we need to call the method of the same name on `I18n` itself. Certainly, I just took off the prefix in this text and re-ran the test, it failed, couldn't resolve `t`, `NoMethodError`.

munckymagik: And I should add, is that ok? It seemed like that was "the way things are done".

rnalexander:
All good!  :+1: :shipit:

munckymagik:
Thanks

PR #322, opened 2015-10-28
Create a rake report to audit the index against the database

munckymagik:
We need to reindex the entire contents of our ElasticSearch index. When we did looked at doing this on the demo server we noticed there were 10 more records in the index than in the database. Not hugely surprising, but we need to work out what is happening here.

This rake task and helper class allow us to work out what's going on.

# Example of the Rake output

```
$ heroku run rake reports:audit_index_csv
Running rake reports:audit_index_csv on mas-rad-demo... up, run.2499
id,frn,name,exists_in_es?,exists_in_db?,publishable?,geocoded?,action
2,,,true,false,,,delete_from_es
4,,,true,false,,,delete_from_es
5,,,true,false,,,delete_from_es
7,,,true,false,,,delete_from_es
8,,,true,false,,,delete_from_es
9,,,true,false,,,delete_from_es
12,,,true,false,,,delete_from_es
40,503220,Redacted Firm Name,false,true,false,true,do_nothing
45,439620,Redacted Firm Name,false,true,false,true,do_nothing
62,501806,Redacted Firm Name,false,true,false,true,do_nothing
71,194713,Redacted Firm Name,false,true,false,true,do_nothing
74,607322,Redacted Firm Name,false,true,false,true,do_nothing
76,194713,Redacted Firm Name,false,true,false,true,do_nothing
109,153420,Redacted Firm Name,false,true,false,true,do_nothing

...
```

# Example console output using the helper class

```
irb(main):001:0> ai = Tasks::AuditIndex.new
=> #<Tasks::AuditIndex:0x007fd7bd2778f0>
irb(main):002:0> ai.all_es_firm_ids.count
=> 2494
irb(main):003:0> ai.all_pg_firm_ids.count
=> 2780
irb(main):004:0> ai.in_es_not_pg
=> [2, 4, 5, 7, 8, 9, 12]
```

### DIFF
--- a/lib/tasks/audit_index.rb
+++ b/lib/tasks/audit_index.rb
@@ -1,0 +1,132 @@
+require 'csv'
+require 'pp'
+require 'io/console'
+
+module Tasks
+  class AuditIndex
+    def all_es_firm_ids
+      @all_es_firm_ids ||= find_all_es_firm_ids
+    end
+
+    def all_pg_firm_ids
+      @all_pg_firm_ids ||= find_all_pg_firm_ids
+    end
+
+    def in_es_not_pg
+      @in_es_not_pg ||= all_es_firm_ids - all_pg_firm_ids
+    end
+
+    def in_pg_not_es
+      @in_pg_not_es ||= all_pg_firm_ids - all_es_firm_ids
+    end
+
+    # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+    def action(o)
+      if o[:publishable?] && o[:geocoded?] && !o[:exists_in_es?]
+        :index
+      elsif o[:publishable?] && !o[:geocoded?]
+        :invalid_firm_address?
+      elsif o[:exists_in_es?] && (!o[:exists_in_db?] || !o[:publishable?])
+        :delete_from_es
+      elsif !o[:exists_in_es?] && !o[:publishable?]
+        :do_nothing
+      else
+        :dunno
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+    def analyse_in_es_not_pg
+      @analyse_in_es_not_pg ||= in_es_not_pg.map do |id|
+        exists_in_db = Firm.exists?(id)
+        firm = Firm.find(id) if exists_in_db
+        res = {
+          id: id,
+          exists_in_es?: true,
+          exists_in_db?: exists_in_db,
+          publishable?: firm.try(&:publishable?),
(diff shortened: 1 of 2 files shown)


lib/tasks/audit_index.rb
```diff
@@ -0,0 +1,132 @@
...
+
+        (analyse_in_es_not_pg + analyse_in_pg_not_es).map do |o|
+          csv << headers.map { |key| o[key] }
+        end
+      end
+    end
+
+    def self.analyse
+      new.tap(&:analyse)
+    end
```
JiggyPete: This class method doesn't seem to be pulling it's weight, if we removed it we could have the following in the rake task
`Tasks::AuditIndex.new.analyse` 

?

munckymagik: Well, I can see your point. I put them in for convenience when using this class directly from the Rails console. They still *feel* convenient to me, so I'd like to keep them `¯\_(ツ)_/¯`

lib/tasks/audit_index.rb
```diff
@@ -0,0 +1,132 @@
...
+    def all_es_firm_ids
+      @all_es_firm_ids ||= find_all_es_firm_ids
+    end
+
+    def all_pg_firm_ids
+      @all_pg_firm_ids ||= find_all_pg_firm_ids
+    end
+
+    def in_es_not_pg
+      @in_es_not_pg ||= all_es_firm_ids - all_pg_firm_ids
```
JiggyPete: I'm not that fussed about this but thought I'd mention.... the memoization isn't needed as this method is only called once.

lib/tasks/audit_index.rb
```diff
@@ -0,0 +1,132 @@
...
+    def find_all_es_firm_ids
+      response = ElasticSearchClient.new.search('firms/_search?size=10000&fields=')
+      res = JSON.parse(response.body)
+      res['hits']['hits'].map do |hit|
+        hit['_id']
+      end.map(&:to_i).sort.freeze
+    end
+
+    def find_all_pg_firm_ids
+      Firm.registered.geocoded.pluck(:id).sort.freeze
```
JiggyPete: I always forget about `:pluck`. This is great!

JiggyPete:
:+1:

lib/tasks/audit_index.rb
munckymagik: Sorry, like I said, this class is intended to be used from the console too. The memoization is for that use-case. Otherwise yes, we wouldn't need to bother.

tomas-stefano:
:+1:

munckymagik:
Thanks

PR #324, opened 2015-11-11
Geocoding offices and advisers with feedback if it fails

munckymagik:
## Summary

This set of changes integrates the work done in mas-rad_core to:

* Make offices geocodable
* Make it so both office and adviser geocoding errors can be reported to the user at the moment they try to save.

Now neither advisers nor offices can be saved without successfully being geocoded. This reduces the number of possible states the model can be in, hopefully making it simpler to reason about how the system should work and what data should be visible in the directory.

## Things to note

VCR has been added as we now need to stub out requests to the geocoding service. Previously when all geocoding and indexing was triggered by `after_commit` hooks no remote requests were being triggered. This was due to the [`DatabaseClearner.strategy` being set to `:transaction`](https://github.com/moneyadviceservice/rad/blob/master/spec/support/database_cleaner.rb#L7) - commits never happened so the callbacks were not triggered. The change now is because geocoding is done as a wrapper around save and update.

Note indexing is still done on an `after_commit` so that doesn't need to be stubbed out.

### Offices

When saving an address whose postcode cannot be geocoded the user gets feedback.

screen shot 2015-11-11 at 12 29 55

### Advisers

Same situation now with advisers too.

screen shot 2015-11-11 at 12 34 30


db/schema.rb
```diff
@@ -11,10 +11,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151102113518) do
+ActiveRecord::Schema.define(version: 20151103161642) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pg_stat_statements"
```
munckymagik: Damn it this guy has crept in again :(

Tr4pSt3R:
Syntax-wise everything is sound and good. If `Address` is now synonymous to `Postcode` you could think about making changes to reflect this fact. 
:+1:

mconnell:
:+1:

munckymagik:
@Tr4pSt3R thanks yes I have an idea to sort that out later.

Tr4pSt3R:
@munckymagik bless!

PR #327, opened 2015-11-18
FCA Import details moved to internal wiki

JiggyPete:
@rnalexander @darokel I've fleshed out the details around the FCA import. Due to the sensitive nature of this information I've added it to the internal wiki rather than it being within this repo.

As you can see below the README that contained all the info now links to the wiki.


spencer:
:+1:

rnalexander:
:+1:  (For the record, using the @ name for me on here won't actually cause me to be alerted in any way that I'll see.  You need to do it in Slack to get my attention.  sorry.)

PR #328, opened 2015-11-20
Update report now that firm records don't have an address

munckymagik:
Delegates to main_office if there is one. Also report needed to be
updated to only analyse firms that have a main office. This is the
equivalent to what was there previously.


mconnell:
:+1:

tomas-stefano:
:+1:

Tr4pSt3R:
:+1:

PR #329, opened 2015-11-20
Reports update

munckymagik:
Stats gen updates based on feedback from the RAD admin team.

I’ve updated the reports with new data, which hopefully makes it clearer, or at least more possible to work out what things mean.

There is now more information in the `adhoc:firm_stats` report. Note, it’s a little unclear what registered actually means after the last few months of changes, so I’ve defined it based on what the code does (see the note in brackets) to try to clear up any confusion.

The by-country report is now generated based on all firms that appear on the directory, which in the code refer to as “published”. The numbers now look consistent with the report generated last month.

### DIFF
--- a/lib/reports/by_country_report.rb
+++ b/lib/reports/by_country_report.rb
@@ -2,21 +2,20 @@ require 'net/http'
 
 class ByCountryReport
   def initialize
-    @registered_firms = Firm.registered.select { |f| f.main_office.present? }
+    @published_firms = Firm.registered.select(&:publishable?)
     @lookup = {}
     generate_lookup
   end
 
   def firm_report
-    firms = @registered_firms.reject { |f| f.parent_id.nil? } # don't include trading names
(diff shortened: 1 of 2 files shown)


urfolomeus:
:+1: lovely

rnalexander:
:+1: Although I can't help but wonder if this is some new reporting requirement in disguise.  Did a stakeholder ask for this explicitly or do you just get bugged about this often enough that you decided to make a quick way to answer?

munckymagik:
@rnalexander thanks. Actually the requirements from the stakeholder haven't changed, the existing statistics are what they want. But I've personally tried to surface a bit more at this point - after discussing with the stakeholder I realised there may have been cause for confusion. She cuts down the output before passing up.

With the by country report it just wasn't answering the same question as it did after the first time we ran it, most likely I didn't get things quite right in #328.

PR #331, opened 2016-02-08
Add advisers without a reference number to a firm

cazrin:
# Summary

Spec from Trello card:

> In some cases, we need to include advisers in the directory who do not have an adviser reference number for us to cross-check against the FCA Register.
>
> This is needed to be able to include equity release advisers who do not have a FCA adviser reference number.
>
> This does not need to be represented in self-service only manual entry by MAS.  This gives us the security that only qualified individuals are being added to directory.

There is a corresponding PR on `mas-rad_core` that this is dependent upon. https://github.com/moneyadviceservice/mas-rad_core/pull/138

# Screen Shots

## Button added to firm#show template

screen shot 2016-02-10 at 14 16 59

## New form added

screencapture-localhost-3000-admin-firms-153-advisers-new-1455113806884

### DIFF
--- a/app/views/admin/firms/show.html.erb
+++ b/app/views/admin/firms/show.html.erb
@@ -56,15 +56,16 @@
 
   <ul>
     <% @firm.advisers.each do |adviser| %>
-      <li><%= link_to adviser.name, admin_adviser_path(adviser) %>, added <%= adviser.created_at.to_s(:short) %></li>
+      <li class='t-adviser'><%= link_to adviser.name, admin_adviser_path(adviser) %>, added <%= adviser.created_at.to_s(:short) %></li>
     <% end %>
   </ul>
 
-  <%= button_to 'Move advisers', new_admin_move_adviser_path,
-                :class => "t-move-advisers", :method => :get %>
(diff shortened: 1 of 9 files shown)


paulanthonywilson:
:+1:

app/views/admin/firms/show.html.erb
```diff
@@ -56,16 +56,17 @@
...
   </ul>
 
-  <%= button_to 'Move advisers', new_admin_move_adviser_path,
-                :class => "t-move-advisers", :method => :get %>
+  <%= link_to 'Move advisers', new_admin_move_adviser_path, class: 'btn btn-default t-move-advisers', method: :get %>
 <% else %>
   <p>No advisers registered.</p>
 <% end %>
 
+<%= link_to 'New adviser', new_admin_firm_adviser_path(@firm), class: 'btn btn-success t-new-adviser', method: :get %>
```
munckymagik: Maybe the link text should make it clearer what this link is for and why we have this in addition to the normal self-service flow for adding advisers?

> New adviser (without an FCA number)

or 

> New equity release adviser

tomas-stefano:
:+1:

PR #332, opened 2016-02-09
Remove asterisk from website address label

cazrin:
Removing the asterisk as the website address isn't actually a required field.


tomas-stefano:
:+1:

johnplayer1982:
:+1:

PR #334, opened 2016-02-15
Adding new phantomjs pre-req and requirejs packages

urfolomeus:
Whenever I install RAD fresh I get a `missing requirejs` error, so I've explicitly added it to the project manifest. Also PhantomJS was complain about a missing pre-req, so I added that too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
     "karma-spec-reporter": "0.0.13",
     "mocha": "~1.15.1",
     "phantomjs": ">= 1.9.7-1",
+    "phantomjs-prebuilt": ">= 1.9",
+    "requirejs": "2.1.22",
     "sinon": "^1.9.0"
   }
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,6 +24,8 @@
     "karma-spec-reporter": "0.0.13",
     "mocha": "~1.15.1",
     "phantomjs": ">= 1.9.7-1",
+    "phantomjs-prebuilt": ">= 1.9",
```
munckymagik: Can we go for Phantom >=2? Seems like this might be the moment to try?

urfolomeus: It seems that we can. :)

Not only that but it removes the requirement to manually add the `-prebuilt` package too. \o/

urfolomeus:
@munckymagik OK it would seem that the best way to do this is to upgrade the karma-phantomjs-runner instead and have it install phantomjs. Need to manually add the -prebuilt though.

Tr4pSt3R:
:+1:

munckymagik:
Checked out, installed and tests run fine. :+1:

PR #335, opened 2016-02-15
Display principal email and telephone number in admin screens

munckymagik:
# Why?

Previously the principal email was not shown anywhere in the admin screens, leaving the admin team with no choice but to navigate all the way into the self-service area to find it. So now it is displayed.

For maximum convenience the principal phone and email are now also shown on the admin firm screens too.

# Screen shots

## Principal 'show' screen

screen shot 2016-02-15 at 14 27 39

## Firm 'show' screen

screen shot 2016-02-15 at 14 27 59

### DIFF
--- a/app/views/admin/firms/show.html.erb
+++ b/app/views/admin/firms/show.html.erb
@@ -8,6 +8,10 @@
     <p>
       <strong>Principal:</strong>
       <%= link_to "#{@firm.principal.first_name} #{@firm.principal.last_name}", admin_principal_path(@firm.principal) %>
+      <ul>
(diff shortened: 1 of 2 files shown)


urfolomeus:
:+1:

slaywell:
Would it be worth having the phone number in a `<a href="tel:" >` tag, like we do with other phone numbers on the MAS website?

munckymagik:
@slaywell good point. I'm not sure. I'll ask Sandrine if she calls from her PC or by desk phone. I suspect it's the latter.

slaywell:
:+1: for this PR, don't need the link on the telephone no as this is non-public-facing.

PR #338, opened 2016-02-17
Report automation: create daily snapshots, show snapshot details in admin

cazrin:
Related PR: https://github.com/moneyadviceservice/mas-rad_core/pull/142

Uses the new `Snapshot` model from `mas-rad_core`. Snapshot captures metrics (queries). This change to `rad` is a way of viewing each Snapshot, as well as setting up the daily job for creating a new Snapshot.


app/views/admin/metrics/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <tr>
+      <th>Date</th>
+      <th></th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% @snapshots.each do |snapshot| %>
+      <tr class="t-snapshot-table-row">
+        <td style="vertical-align: middle"><%= snapshot.created_at.strftime('%e %B %Y') %></td>
```
mconnell: inline style... maybe want to do something like:

`<td class="table__cell--middle">...</td>`

Or some other BEM name.

mconnell:
Maybe a note about running sidekiq in the README.

mconnell:
Other than small comments :+1:

app/views/layouts/admin.html.erb
```diff
@@ -26,6 +26,7 @@
         <li class="<%= 'active' if current_page?(controller: '/admin/firms') %>"><%= link_to 'Firms', admin_firms_path %></li>
         <li class="<%= 'active' if current_page?(controller: '/admin/advisers') %>"><%= link_to 'Advisers', admin_advisers_path %></li>
         <li class="<%= 'active' if current_page?(controller: '/admin/principals') %>"><%= link_to 'Principals', admin_principals_path %></li>
+        <li class="<%= 'active' if current_page?(controller: '/admin/metrics') %>"><%= link_to 'Metrics', admin_metrics_path %></li>
```
Tr4pSt3R: it will be great to have this in a partial. (minor 2pence)

Tr4pSt3R:
:+1: :+1: :+1:

app/views/layouts/admin.html.erb
cazrin: I think I'll probably just replace this with https://github.com/comfy/active_link_to as it solves the problem nicely.

app/controllers/admin/metrics_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+     :advisers_with_qualification_in_pension_transfer,
+     :advisers_with_qualification_in_equity_release,
+     :advisers_with_qualification_in_long_term_care_planning,
+     :advisers_with_qualification_in_tep, :advisers_with_qualification_in_fcii,
+     :advisers_part_of_personal_finance_society,
+     :advisers_part_of_institute_financial_planning,
+     :advisers_part_of_institute_financial_services,
+     :advisers_part_of_ci_bankers_scotland,
+     :advisers_part_of_ci_securities_and_investments,
+     :advisers_part_of_cfa_institute, :advisers_part_of_chartered_accountants]
```
tomas-stefano: is this the attributes from the model? Are you sure that you can't use `Snapshot.attributes.map(:&to_sym)` or something related?

tomas-stefano: Or use .column_names from AR

tomas-stefano: The advantange is that with this approach you avoid to skip rubocop through the code.

cazrin: I do like the idea of using `.column_names`. The only downside I can think of is that I need to have them all in a specific order (not alphabetical, or whichever way I did the migrations). Is there a better solution?

tomas-stefano: Ok. I can't think a way to do it, since is ordered at random, but I would move all this private methods to a ruby class that handle the conversion.

README.md
```diff
@@ -61,11 +61,15 @@ To start the application:
 
 ```sh
 $ bundle exec rails s -p 5000
+$ bundle exec sidekiq
```
munckymagik: We need to add a note that Redis must be running locally in order to sidekiq

munckymagik:
:+1:

mconnell:
:+1:

tomas-stefano:
:+1:

PR #340, opened 2016-02-19
Prevent Heroku timeout on CSV download

urfolomeus:
Heroku was timing out before the file was downloaded. This was partly
because the timeout was set too low, but mostly because I'm an idiot who
forgot to check for n+1 issues with the data fetch query >.<


tomas-stefano:
:+1:

rnalexander:
:+1: :shipit:

urfolomeus:
Thanks @rnalexander @tomas-stefano :)

PR #342, opened 2016-02-19
[DEV APPROVED] Remove "Other" accreditation: modify seeds file

cazrin:
Related PRs: https://github.com/moneyadviceservice/rad_consumer/pull/271 and https://github.com/moneyadviceservice/mas-rad_core/pull/145

* Removed the "Other" accreditation from the seeds file
* Use latest version of mas-rad_core
* Run `rake db:migrate`

Once the migrations are run I think a `rake firms:index` is needed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'dough-ruby',
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
-gem 'mas-rad_core', '0.0.101'
+gem 'mas-rad_core', '0.0.102'
 gem 'oga'
 gem 'pg'
 gem 'rails_email_validator'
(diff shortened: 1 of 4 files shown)


urfolomeus:
:+1:

munckymagik:
This migration should (counter-intuitively) go into mas-rad_core I think. The seeds file should not really be in RAD, but it is.

The only migrations in RAD should be those that affect only RAD (i.e. the Devise/user related stuff).

I think it would be really good to at least maintain the separation we have. This change affects both rad_consumer and rad so it should happen in mas-rad_core.

cazrin:
Good point @munckymagik, I'll move it over to mas-rad_core. The reason I did it in RAD initially is because of the seeds file and that RAD "owns the database", but it makes sense that all things data should exist in mas-rad_core.

munckymagik:
:+1:

rnalexander:
What's going on with this?  It seems not-ideal to be pointing at a branch of our own code library.

cazrin:
I'm waiting on the gem PR to get a :+1: before I point to the non-branch version of the gem. I was told this was common practice and have been doing it for the other PRs without any issues.

rnalexander:
Cool cool :+1:

PR #346, opened 2016-02-23
Fix occasionally failing test

urfolomeus:
The sort was blowing up very occasionally because the `sort_by` sort does not
sort things in the same order as `order` does if there are commas in the
name.

```
sorted_names = assigned_names.sort_by(&:downcase)

expected: ["Buckridge-Beier", "Hessel LLC", "Hessel, Kovacek and Marvin", "Labadie, Johnson and Glover", ...]
     got: ["Buckridge-Beier", "Hessel, Kovacek and Marvin", "Hessel LLC", "Labadie, Johnson and Glover", ...]
```


cazrin:
:+1:

Tr4pSt3R:
:+1:

PR #347, opened 2016-02-24
Registration - redirects

urfolomeus:
The way the existing site redirects users is confusing. For example, successfully editing an Adviser lands you back on the Edit page. This makes the following adjustments:

1) Successfully creating/updating an Adviser redirects to the Adviser index page.
2) Successfully creating/updating an Office redirects to the Office index page.
3) The Firm still redirects to the Edit page as there is another change coming through which requires this behaviour.


munckymagik:
:+1:

Tr4pSt3R:
:+1:

urfolomeus:
Thanks @munckymagik @Tr4pSt3R :)

PR #348, opened 2016-02-24
Registration - new onboarding message

urfolomeus:
The original onboarding message was confusing as it looked like a flash message and would sometimes visually conflict with other flash messages being displayed. The onboarding message is now just a flat grey panel and contains more descriptive text.

We also now only show this message on the Firm index page as we have a separate message (being built on another card) on the other pages.

screenshot 2016-02-24 14 59 11


Tr4pSt3R:
:+1:

cazrin:
:+1:

PR #355, opened 2016-02-26
Making colour of exclamation status more accessible

urfolomeus:
The colours that we were using originally didn't have enough contrast. The new colour scheme meets accessibility requirements.

## Before

screenshot 2016-02-26 08 57 17

## After

screenshot 2016-02-26 08 56 58


slaywell:
:+1:

app/assets/stylesheets/components/_status.scss
```diff
@@ -6,7 +6,12 @@
 }
 
 .status__icon--exclamation {
-  fill: $color-yellow-dark;
+  background-color: $color-yellow-light;
+  border-radius: 100%;
```
cazrin: Interesting that 100% worked, I'm sure this is usually meant to be set to 50% to create the circle.

urfolomeus: Yeah that was my understanding too, but I was getting an oblong on mobile for some reason. Might have been something else causing that mind you. My brain is full of the cold :)

cazrin:
Did you look into making the tick/exclamation the same width? Taking into account the padding. Just to line them up horizontally in the status panel.

urfolomeus:
Good shout. Will sort that now.

urfolomeus:
@cazrin implemented your suggestions. @slaywell I've used the same folder structure now as other components that we have that have their own settings. Can you re-thumb if you're ok with this please?

cazrin:
:+1: 

I'm guessing that weirdly 50% border-radius started working again? :)

urfolomeus:
Yeah. I'll blame it on the Lemsip ;)

slaywell:
I think this SCSS patter is RAD-specific, seems a bit much to have a separate folder with SCSS variables for components. But... it is consistent with the rest of RAD, so :+1:

urfolomeus:
It was something that Ben Barnett asked us to introduce IIRC, but I might be misunderstanding its use/purpose. As you say, at least it's consistent :)

Thanks @slaywell.

PR #359, opened 2016-02-26
Provide a way for MAS admins to delete an entire firm

munckymagik:
# What?

Adds a button to the private MAS admin UI that allows deletion of a firm and all related trading names from the directory.

**Note** Counter-intuitively this is done by deleting the principal as the principal object "owns" the parent firm and trading names so deleting that cascades through correctly.

# Why?

Whenever firms disappear from the FCA register or ask to be taken off of the directory we need a way to delete them. IIRC the frequency has been about 2 or 3 a month.

Until now, firm deletion has been a BAU task done on the Rails console whenever it was necessary.

# Screenshots

screen shot 2016-02-26 at 15 24 22

screen shot 2016-02-26 at 15 18 47

### DIFF
--- a/app/controllers/admin/principals_controller.rb
+++ b/app/controllers/admin/principals_controller.rb
@@ -10,10 +10,15 @@ class Admin::PrincipalsController < Admin::ApplicationController
 
   def destroy
     user = User.find_by(principal: principal)
+
+    message = "Successfully deleted #{principal.full_name}, " \
+              "#{principal.firm.registered_name} and " \
+              "#{principal.firm.trading_names.count} related trading names."
+
     user.principal.destroy
     user.destroy
 
-    redirect_to admin_principals_path
+    redirect_to admin_principals_path, notice: message
   end
 
   private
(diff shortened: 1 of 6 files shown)


munckymagik:
@rnalexander you may like this.

Tr4pSt3R:
Woah this is a dangerous button!

app/controllers/admin/principals_controller.rb
```diff
@@ -8,6 +8,19 @@ def show
     @principal = principal
   end
 
+  def destroy
```
Tr4pSt3R: Hey @munckymagik, Is there sufficient authorisation check to ensure that it is only the admin who can destroy a principle please?

munckymagik: Sure the entire `/admin` route is wrapped in HTTP auth:

https://github.com/moneyadviceservice/rad/blob/master/config/routes.rb#L81

screen shot 2016-02-26 at 15 54 15

Tr4pSt3R: @munckymagik that looks good

Tr4pSt3R:
:+1:

cazrin:
:+1:

munckymagik:
Thanks for the review @Tr4pSt3R @cazrin

PR #361, opened 2016-03-01
Don't show validation errors until registration form submitted

jensljungblad:
screen shot 2016-03-01 at 11 31 31

We should not call `.valid?` within the view because this will cause validation errors to display before the form has been submitted.

Instead we should use the existing `Firm#registered?` as this will tell us whether the firm record has been completed without triggering validation.


JiggyPete:
:+1:

rnalexander:
:+1:

PR #365, opened 2016-03-02
Fix svg icons for IE

munckymagik:
IE11 does not support loading an SVG spritesheet as an external source.

# Before

screen shot 2016-03-02 at 12 24 53

# After

screen shot 2016-03-02 at 14 27 46


JiggyPete:
:+1:

slaywell:
:+1:

munckymagik:
Thanks both.

PR #367, opened 2016-03-03
[DEV APPROVED] Filter duplicate trading names

munckymagik:
# What?

We have now added an additional filtering step to the FCA data import scripts so we only include unique and active trading names.

# Why?

The data file from the FCA that defines the trading names registered for a firm can contain multiple entries for the same name. Each entry has a 'start date' and an 'end date' column. Entries with an end date are now filtered out. We include only entries that have an empty end date.

# P.S.

Even after doing this we found roughly 20 trading names (out of nearly 70,000) that had more than one entry without an end date. As we only import the FCA number and the name itself, any other differences are irrelevant. So for these we take just one copy of each of these names (see `is_unique_trading_name`).


mconnell:
:+1:

rnalexander:
:+1:

PR #373, opened 2016-03-08
Allow out of date names to be updated

mconnell:
We added a a report to the admin section that was previously run to highlight companies where the name  used within the directory, is different to the current FCA records.

This PR updates the report to include the functionality to, on a case-by-case basis, replace the registered name displayed in the directory with the latest name according to the FCA.


munckymagik:
:+1:

Tr4pSt3R:
:+1:

PR #375, opened 2016-03-09
Ensure header displays in IE8

mconnell:
The background for the header wasn't appearing in IE8, so it gave the impression the header didn't exist at all. Highlighting the header in the browser reveals that the content is there. However, with it being white, it didn't appear to be visible.

screen shot 2016-03-09 at 10 23 52

The issue appears to be that IE8 was rendering the `<header>` element with a `display: inline` property, regardless of style being declared `header { display: block; } `. I've not investigated beyond resolving the issue, but I suspect there is perhaps an issue with applying styles to HTML elements which are not natively supported by IE8.

Declaring `display: block` on the class used by the element appears to resolve the issue.

screen shot 2016-03-09 at 10 24 07

### DIFF
--- a/app/assets/stylesheets/layouts/_header.scss
+++ b/app/assets/stylesheets/layouts/_header.scss
@@ -1,5 +1,6 @@
 .l-header {
   height: $baseline-unit*12;
+  display: block;
 }
 
 .l-header .mas-logo {


johnplayer1982:
Pulled this locally and tested on Browserstack :+1:

slaywell:
:+1:

mconnell:
Thanks @johnplayer1982 + @slaywell

PR #376, opened 2016-03-09
Ensure new advisers without reference numbers are geocoded when created

munckymagik:
This fixes an issue we noted where creating new advisers with this new admin-only form added them to the Postgres database but they didn't show up in ElasticSearch.


Tr4pSt3R:
Looks good
:+1:

paulanthonywilson:
:+1:

PR #377, opened 2016-03-09
Change move advisers button color

jensljungblad:
This PR is to change the move advisers button colour to match the 'Sign in as...' button.

'Sign in as...' is an input button with the browsers default styling.
Previously 'Move advisers' was a bootstrap styled link, so we flipped that to `button_to` to look the same as 'Sign in as...'

Before:
screen shot 2016-03-09 at 16 06 50

After:
screen_shot_2016-03-09_at_15 53 27_360

### DIFF
--- a/app/assets/stylesheets/admin.css.scss
+++ b/app/assets/stylesheets/admin.css.scss
@@ -65,3 +65,7 @@ code {
 .adviser__actions__edit {
   @include make-xs-column(6);
 }
+
+.firm__move-advisers {
+  display: inline;
+}
--- a/app/views/admin/firms/show.html.erb
+++ b/app/views/admin/firms/show.html.erb
@@ -64,7 +64,7 @@
     <% end %>
   </ul>
 
-  <%= link_to 'Move advisers', new_admin_move_adviser_path, class: 'btn btn-default t-move-advisers', method: :get %>
+  <%= button_to 'Move advisers', new_admin_move_adviser_path, class: 't-move-advisers', method: :get, form_class: 'firm__move-advisers' %>
 <% else %>
   <p>No advisers registered.</p>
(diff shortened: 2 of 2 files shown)


munckymagik:
I think these very similarly named commits need to be squashed together and we should explain why we're deliberately not using the Bootstrap provided style.

JiggyPete:
@munckymagik good point

JiggyPete:
@munckymagik does the new description help?

munckymagik:
Sorry my point was, the end-user (and stakeholder) specifically requested this rather than the Bootstrap style.

munckymagik:
:+1:

slaywell:
Seems like a strange request. Do you have to log in to 'move advisers', or is there some other reason that the 'log in' button styling is more appropriate? I preferred it when the buttons were the same size... But not a major issue.

slaywell:
:+1:

munckymagik:
@slaywell No need to login for 'move advisers'. Our stakeholder wants to be able to visually differentiate the buttons in exactly this way. This was her specific request. As we're not the end user here and this isn't consumer facing we're taking a step back to give someone what they need to make their life easier.

JiggyPete:
@slaywell you don't have to log in to 'move advisers'. This is on an admin page, the admin user can move advisers between firms, they can also 'Sign in as <firm user>' to be able to do everything the firm user can.

I agree about the buttons, we did put forward an example of that, however Sandrine (who is the main admin user) asked for what you see in the description.

slaywell:
Sorry guys I replaced my comment with a thumb whilst you were responding as I realised this is a MAS admin only page!

JiggyPete:
@slaywell thanks

PR #378, opened 2016-03-09
add cancel/save to top of form

mconnell:
There has been feedback from advisers that the save and cancel buttons are easy to miss because they are too far down the page. So advisers tend to navigate away from the page without saving if they don't happen to scroll all the way to the bottom.

This PR simply duplicates the save button to the top of the form to make it a little harder to miss.


munckymagik:
:+1:

rnalexander:
:+1:

PR #379, opened 2016-03-09
Exclude advisers added outside of the ref checking process from invalid adviser report

mconnell:
There is functionality to add advisers to the application without having to verify a reference number for the person. These records should not appear in the invalid advisers report.

__Before: 9 records, 3 are 'NOREF'__
screen shot 2016-03-09 at 15 57 27

__After: 6 records__
screen shot 2016-03-09 at 15 57 42


munckymagik:
:+1:

rnalexander:
:+1:

PR #380, opened 2016-03-09
Replace Rackspace CDN urls with Azure CDN urls

mconnell:
Updates the CDN urls we're using from Rackspace to Azure


Tr4pSt3R:
Yup! all the replacement links are valid.
:+1:

munckymagik:
:+1:

PR #383, opened 2016-03-14
[DEV APPROVED]Bug Fix: Controller couldn't find Admin::Reports::FirmsAdvisers

JiggyPete:
The class `Reports::FirmsAdvisers` is defined in `/lib/reports`.  Previously it was used in `Admin::FirmsController` without incident.

Recently the module `Admin::Reports` was introduced, causing `Reports::FirmsAdvisers` to be resolved to `Admin::Reports::FirmsAdvisers` within `Admin::FirmsController`.

This fixes the issue by explicitly specifying the `Reports` module in the root namespace, ie with `::Reports::FirmsAdvisers`.

### DIFF
--- a/app/controllers/admin/firms_controller.rb
+++ b/app/controllers/admin/firms_controller.rb
@@ -15,7 +15,7 @@ class Admin::FirmsController < Admin::ApplicationController
   end
 
   def adviser_report
-    data = Reports::FirmsAdvisers.data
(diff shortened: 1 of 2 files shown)


paulanthonywilson:
:+1: (had a shot at adding more explanation to the  PR description.)

rnalexander:
:+1:

PR #386, opened 2016-03-23
Backend App: Upgrade to dough 5.12 and integrate latest frontend CSS from the CDN

munckymagik:
Brings this app up to date with the frontend and the latest Dough.

* Swaps to using tags to track versions of Dough so it's easy to see the exact version we are referencing.
* Yeast is still configured to track master in bower.json.erb so all fresh deployments naturally get the latest version anyway.
* Updated the URLs for assets we import from frontend via the CDN by looking at the page source for http://www.moneyadviceservice.org.uk today.
* Made changes necessary to integrate the new [mobile sticky header](https://github.com/moneyadviceservice/frontend/pull/1365).

Broadly tested using BrowserStack on:

* IE8, IE11
* iPhone 6, iPad
* Galaxy S6
* FF latest
* Safari latest
* Chrome latest

screen shot 2016-03-23 at 18 26 19

The related change in the consumer application is here: https://github.com/moneyadviceservice/rad_consumer/pull/292

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,10 +17,12 @@ gem 'bowndler', github: 'moneyadviceservice/bowndler'
 gem 'devise'
 gem 'devise_invitable'
 gem 'devise_security_extension'
+# Dough assets are loaded from a CDN instead of from the Gem. Do make sure that
+# the CDN version is the same as the Gem version.
 gem 'dough-ruby',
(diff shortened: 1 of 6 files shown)


slaywell:
:+1:

JiggyPete:
:+1:

PR #391, opened 2016-04-26
Use incoming for new files

misaka:
Inspired by a problem where there were 2 sets of updates in archives causing much pain and anguish. And an error.


jiannis:
👍

munckymagik:
👍

PR #393, opened 2016-07-11
7396 principal and user editing

jongilbraith:
Admins currently can only modify such information by using the 'login as principal' functionality, which works for most things, but will not permit them to reset their password.

This PR resolves this issue by adding a real back for managing a principal's main directory details and their login credentials (the `User` model).

Additionally I found that a bunch of features had their step methods defined in the global scope, causing overlap, so I fixed those up. Plus a few other bits of tidying.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ group :test, :development do
   gem 'factory_girl_rails'
   gem 'faker'
   gem 'pry-rails'
+  gem 'pry-byebug'
   gem 'rspec-rails'
   gem 'rubocop'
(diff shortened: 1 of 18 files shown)


munckymagik:
👍

tomas-stefano:
👍

rnalexander:
:+1: :shipit:

jongilbraith:
Thought it would be good to give Sandrine a quick demo of this, but now waiting for someone else to be in so I can demo it to them as well.

PR #397, opened 2016-09-28
[BUG] rename column registered_name to name on the lookup_subsidiaries table

jriga:
**Description**
The FCA import script uses temporary table to load the data. However there was a mistake on the lookup_subsidiaries table schema.

**Solution**
rename registered_name to name on the temp lookup_subsidiaries table


jongilbraith:
👍

PR #398, opened 2016-10-03
macOS setup automation via suited

norm:
Assuming you are setting up a new Mac with github.com/norm/suited,
these files will install the dependencies for the project.

(They can also be used to manually install dependencies if you're not
using suited.)


jongilbraith:
👍

jongilbraith:
👍

norm:
@jriga good call, added :)

rnalexander:
👍 :shipit:

PR #400, opened 2016-11-18
[WIP] Create postgres's postgres user idempotently

norm:
Running this script twice would fail because the user had already been
created. No longer the case.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,6 +1,7 @@
 set -e
 
-createuser -ds postgres
+psql -c '\du' postgres | grep -q ' postgres .*Superuser' \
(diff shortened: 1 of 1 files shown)


jongilbraith:
👍

pmcostello:
hmm not sure it's a good idea to create the postgres role as a superuser without a secure password, otherwise its a potential easily exploitable attack vector on dev machines, especially as some of the local app db's contain prod data i.e. rad can we agree a default mas_postgres user/role i.e maspgsql?

norm:
@pmcostello Sure, let's discuss it.

paulanthonywilson:
Very old.

PR #406, opened 2017-05-23
[DEV APPROVED] TP#8070: Adds ability to download registered advisers as a CSV file

Tr4pSt3R:
[Card TP#8070: Registered Advisers Export List](https://moneyadviceservice.tpondemand.com/entity/8070)

yj6onbvdbp

### DIFF
--- a/app/controllers/admin/reports/registered_adviser_controller.rb
+++ b/app/controllers/admin/reports/registered_adviser_controller.rb
@@ -1,0 +1,21 @@
+module Admin
+  module Reports
+    class RegisteredAdviserController < Admin::ApplicationController
+      respond_to :html, :csv
+
+      def index
+        advisers = Adviser.all
+        @registered_advisers = advisers.page(params[:page])
+        @registered_advisers_list = AdviserListCsv.new(advisers)
+
+        respond_with(@registered_advisers_list, filename: filename)
+      end
+
+      private
+
+      def filename
+        "registered_advisers_#{Time.zone.today}"
+      end
+    end
+  end
+end
--- a/app/views/admin/reports/registered_adviser/index.html.erb
+++ b/app/views/admin/reports/registered_adviser/index.html.erb
@@ -1,0 +1,40 @@
+<h2>Registered Advisers</h2>
+
+<p>
+  This report shows all advisers whose FCA status is still 'Active'.
+</p>
+
+<p>
+  The <a href="https://register.fca.org.uk/">Financial Services Register</a>
+  can be searched to determine exactly what status change has occurred.
+</p>
+
+<% if @registered_advisers.present? %>
+  <p>
+    <%= link_to 'Download as CSV',
+          admin_reports_registered_adviser_index_path(format: :csv),
+          class: 'btn btn-primary' %>
+  </p>
+<% end %>
+
+<table class="table table-bordered">
+  <tr>
+    <th>Ref Number</th>
+    <th>Name</th>
+    <th>Firm</th>
+  </tr>
+  <% if @registered_advisers.present? %>
+    <% @registered_advisers.each do |adviser| %>
+      <tr>
+        <td><%= adviser.reference_number %></td>
+        <td><%= link_to adviser.name, admin_adviser_path(adviser) %></td>
+        <td><%= link_to adviser.firm.registered_name, admin_firm_path(adviser.firm) %></td>
+      </tr>
+    <% end %>
+  <% else %>
+    <tr>
+      <td colspan="3">No advisers to display</td>
+    </tr>
+  <% end %>
+  <%= paginate @registered_advisers %>
+</table>
--- a/config/initializers/csv_renderer.rb
+++ b/config/initializers/csv_renderer.rb
@@ -1,0 +1,9 @@
+require 'csv'
+
+ActionController.add_renderer :csv do |csv, options|
+  filename = options[:filename] || options[:template]
+  data = csv.to_csv(options)
+  send_data data,
+            type: Mime::CSV,
+            disposition: "attachment; filename=#{filename}.csv"
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -82,6 +82,7 @@ Rails.application.routes.draw do
         end
       end
       resource :inactive_adviser, only: [:show]
+      resources :registered_adviser, only: [:index]
       resource :inactive_firm, only: [:show]
       resource :inactive_trading_name, only: [:show]
       resources :out_of_date_firms, only: [:index, :update]
--- a/spec/controllers/admin/reports/registered_adviser_controller_spec.rb
+++ b/spec/controllers/admin/reports/registered_adviser_controller_spec.rb
@@ -1,0 +1,13 @@
+RSpec.describe Admin::Reports::RegisteredAdviserController, type: :controller do
+  describe '#filename' do
+    context 'timestamp' do
+      let(:date) { Date.new(2017, 05, 29) }
+
+      it 'returns filename with timestamp included' do
+        Timecop.freeze(date)
+        expect(controller.send(:filename)).to eq('registered_advisers_2017-05-29')
+        Timecop.return
+      end
+    end
+  end
+end
--- a/spec/features/admin/export_advisers_list_spec.rb
+++ b/spec/features/admin/export_advisers_list_spec.rb
@@ -1,0 +1,96 @@
+# Feature: Advisors list export
+#  As the person maintaining the RAD database.
+#  In order to get rid of unwanted advisors.
+#  I need to be able to cross reference a large list (4k) of unauthorised advisers against records in the RAD database.
+#  The list needs to be cleared up so customers do not get sent to unapproved advisers and RAD maintain it's credibility.
+#
+
+RSpec.feature 'registered advisors report' do
+  let(:admin_page) { Admin::IndexPage.new }
+  let(:registered_adviser_show_page) { Admin::RegisteredAdviserShowPage.new }
+
+  before do
+    given_i_am_a_fully_registered_principal_user
+    and_i_am_logged_in_as_an_admin
+  end
+
+  scenario 'no registered advisers' do
+    and_there_are_no_registered_advisers
+    when_i_visit_the_registered_adviser_page
+    then_i_should_not_see_any_registered_advisers
+  end
+
+  scenario 'has registered advisers' do
+    and_there_exists_registered_advisors
+    when_i_visit_the_registered_adviser_page
+    then_i_should_see_all_registered_advisers
+  end
+
+  scenario 'blocks download as CSV' do
+    and_there_are_no_registered_advisers
+    when_i_visit_the_registered_adviser_page
+    then_i_should_not_be_able_to_download
+  end
+
+  scenario 'permits download as CSV' do
+    Timecop.freeze Date.new(2017, 05, 30) do
+      and_there_exists_registered_advisors
+      when_i_visit_the_registered_adviser_page
+      and_click_on_download_as_csv_button
+      then_i_should_have_a_file_with_all_registered_advisers
+    end
+  end
+
+  def given_i_am_a_fully_registered_principal_user
+    @principal = FactoryGirl.create :principal
+    @user = FactoryGirl.create :user, principal: @principal
+  end
+
+  def and_i_am_logged_in_as_an_admin
+    login_as(@user, scope: :admin)
+    admin_page.load
+    expect(page).to have_content 'RAD Administration'
+  end
+
+  def when_i_visit_the_registered_adviser_page
+    registered_adviser_show_page.load
+  end
+
+  def and_there_are_no_registered_advisers
+    Adviser.delete_all
+  end
+
+  def then_i_should_not_see_any_registered_advisers
+    expect(registered_adviser_show_page).to have_content('No advisers to display')
+  end
+
+  def and_there_exists_registered_advisors
+    firm = FactoryGirl.create(:firm, registered_name: 'James Andrews Investment PLC')
+    FactoryGirl.create(:adviser,
+                       name: 'James Andrews',
+                       reference_number: 'ABC12345',
+                       firm: firm)
+  end
+
+  def then_i_should_see_all_registered_advisers
+    expect(registered_adviser_show_page).to have_content('ABC12345')
+    expect(registered_adviser_show_page).to have_content('James Andrews')
+    expect(registered_adviser_show_page).to have_content('James Andrews Investment PLC')
+  end
+
+  def and_click_on_download_as_csv_button
+    click_on 'Download as CSV'
+  end
+
+  def then_i_should_have_a_file_with_all_registered_advisers
+    filename = 'registered_advisers_2017-05-30.csv'
+
+    expect(registered_adviser_show_page.response_headers['Content-Type']).to eq('text/csv')
+    expect(registered_adviser_show_page.response_headers['Content-Disposition']).to eq("attachment; filename=#{filename}")
+    expect(registered_adviser_show_page).to_not have_content('Registered Advisers')
+  end
+
+  def then_i_should_not_be_able_to_download
+    expect(registered_adviser_show_page).to_not have_content('Download as CSV')
+  end
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -456,4 +456,4 @@ RUBY VERSION
    ruby 2.2.2p95
 
 BUNDLED WITH
-   1.12.5
+   1.15.0
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -31,6 +31,7 @@
           <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">Reports<span class="caret"></span></a>
           <ul class="dropdown-menu" role="menu">
             <%= active_link_to 'Metrics', admin_reports_metrics_path, wrap_tag: 'li' %>
+            <%= active_link_to 'Registered Advisers', admin_reports_registered_adviser_index_path, wrap_tag: 'li' %>
             <%= active_link_to 'Inactive Advisers', admin_reports_inactive_adviser_path, wrap_tag: 'li' %>
             <%= active_link_to 'Inactive Firms', admin_reports_inactive_firm_path, wrap_tag: 'li' %>
             <%= active_link_to 'Inactive Trading Names', admin_reports_inactive_trading_name_path, wrap_tag: 'li' %>
--- a/lib/adviser_list_csv.rb
+++ b/lib/adviser_list_csv.rb
@@ -1,0 +1,15 @@
+class AdviserListCsv
+  def initialize(advisers)
+    @advisers = advisers
+  end
+
+  def to_csv(_ = {})
+    CSV.generate do |csv|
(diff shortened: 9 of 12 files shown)


app/controllers/admin/reports/registered_adviser_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+#  def self.to_csv(_ = {})
+#    CSV.generate do |csv|
+#      csv << ['Ref. Number', 'Name', 'Firm']
+
+#      @advisers.each do |adviser|
+#        csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
+#      end
+#    end
+#  end
+# end
```
margOnline: Can all this commented out code be removed.

app/views/admin/reports/registered_adviser/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <td><%= link_to adviser.name, admin_adviser_path(adviser) %></td>
+        <td><%= link_to adviser.firm.registered_name, admin_firm_path(adviser.firm) %></td>
+      </tr>
+    <% end %>
+  <% else %>
+    <tr>
+      <td colspan="3">No advisers to display</td>
+    </tr>
+  <% end %>
+	<%= paginate @registered_advisers %>
```
margOnline: fix spacing

lib/ext/adviser.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module AdviserExtension
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def to_csv(_ = {})
+      CSV.generate do |csv|
+        csv << ['Ref. Number', 'Name', 'Firm']
+
+        find_each do |adviser|
```
margOnline: 👍  on find_each

spec/features/admin/export_advisers_list_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+# The criteria for entry to the Money Advice Service directory was decided by
+# an independent panel consisting of both consumer and advisers representative
+# bodies.
+# We recently found that some of this advisers have been entering our
+# directory through other company that they are creating. These one are
+# potentialy scams.
+#
+# In order to make sure we don't have any of these advisers in our directory
+# (RAD). Therefore, we need to compare the list of advisers registered under
+# St James place on the FCA register with our database.
```
margOnline: I dont think this is the appropriate place for this information. Your tests are nicely worded and explain what's going on well enough.

app/controllers/admin/reports/registered_adviser_controller.rb
tomas-stefano: You can remove this comment lines here.

db/schema.rb
```diff
@@ -15,7 +15,6 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_stat_statements"
```
tomas-stefano: Is this was intended for the csv to work?

lib/ext/adviser.rb
```diff
@@ -0,0 +1,19 @@
...
+          csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
+        end
+      end
+    end
+  end
+end
+
+ActiveRecord::Base.send(:include, AdviserExtension)
```
tomas-stefano: Is there any more benefits to be a module rather than a class that wraps the advisers list?

config/initializers/csv_renderer.rb
```diff
@@ -0,0 +1,9 @@
+require 'ext/adviser'
+
+ActionController.add_renderer :csv do |csv, options|
+  filename = options[:filename] || options[:template]
```
margOnline: Would be nice for the filename to include a timestamp.

lib/ext/adviser.rb
tomas-stefano: If you're using this module only in one class, why include in all active record base? This means that if I call Office.to_csv they assume that the object quacks like an Adviser. I think that's was not the intended.

tomas-stefano: I would add a unit tests too around the to_csv.

app/views/admin/reports/registered_adviser/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+</p>
+
+<p>
+  The <a href="https://register.fca.org.uk/">Financial Services Register</a>
+  can be searched to determine exactly what status change has occurred.
+</p>
+
+<% if @registered_advisers.present? %>
+  <p>
+    <%= link_to 'Download as CSV', admin_reports_registered_adviser_index_path(format: :csv), download: 'registered_advisers.csv', class: 'btn btn-primary' %>
```
margOnline: I'd like to see us move away from long lines, ideally to 80 chars, but this is open to agreement by the team.

tomas-stefano:
I would reword or squash the commits too.

app/controllers/admin/reports/registered_adviser_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module Admin
+  module Reports
+    class RegisteredAdviserController < Admin::ApplicationController
+      respond_to :html, :csv
+
+      def index
+        @registered_advisers = Adviser.all.page(params[:page])
+        respond_with(@registered_advisers)
```
willhall88: It looks to me that `@registered_advisers` is a paginated number of the total. This is fine for the html page, *but* does the csv then only return the same results rather than ALL of the registered advisors? 
Is this the requirements we need and are there any tests to cover this count?

spec/models/adviser_spec.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.describe Adviser, type: :model do
```
willhall88: Perhaps this test should be updated to test `lib/ext/adviser.rb` and more explicitly test the outcome of that method.

lib/ext/adviser.rb
willhall88: Agree with @tomas-stefano, the test - [here](https://github.com/moneyadviceservice/rad/pull/406/files#diff-fcb272b2a5bcd68e29fa8e3e4051789dR1) should probably be updated to more explicitly test this code

spec/models/adviser_spec.rb
willhall88: initial thoughts ...

move `spec/models/adviser_spec.rb` -> `spec/lib/adviser_spec.rb`

- `Advisor` class should respond to csv
- `Office` class should not
- Response of `.to_csv` should return a csv with the headers `'Ref. Number', 'Name', 'Firm'`

tomas-stefano: What's the difference between the Adviser model and the Advisor in the lib?

willhall88: Adviser was originally put in `models/` but then made better sense to have it in `lib/`, just looks like the spec file hasn't been updated accordingly.

config/initializers/csv_renderer.rb
willhall88: ♻️  @margOnline : https://github.com/moneyadviceservice/rad/pull/406/commits/65ba592e05deb5fac5a37900d5cff6c53428d92d

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.describe Admin::Reports::RegisteredAdviserController, type: :controller do
+  describe '#filename' do
```
tomas-stefano: IMHO, the only tests that you should worry about are the ones testing the public interface. Otherwise you may be coupling too tightly to implementation. Or make it the filename method to be public. Or even better, you can move the timestamp call to the renderer object and then your acceptance tests cover it.

lib/ext/adviser.rb
```diff
@@ -0,0 +1,19 @@
...
+
+        find_each do |adviser|
+          csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
+        end
+      end
+    end
+  end
+end
+
+Adviser.send(:include, AdviserExtension)
```
tomas-stefano: Is there a reason to add the include outside of the Adviser class definitions?
Other thing, I don't want to over engineering, but what happens if we need another type of csv for the Adviser model? Since it was added the to_csv to the Adviser model on this extension, the scope of the to_csv, *it's closed for extension*. Can we not use a wrapper on Adviser model instead a module?

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
Tr4pSt3R: Thanks for your feedback @tomas-stefano. First I will address the better option:

```
Or even better, you can move the timestamp call to the renderer object and then your acceptance tests cover it.
```

The timestamped filename warranted moving this feature to the controllers where it can be dynamically generated before being passed to the CSV renderer. If the filename was included as part of the acceptance test it will be awkward and unwieldy to test for past, present and future scenarios. 

```
scenario 'permits download as CSV' do
    and_there_exists_registered_advisors
    when_i_visit_the_registered_adviser_page
    and_click_on_download_as_csv_button
    then_i_should_have_a_file_with_all_registered_advisers
  end
```

The scenario above will have to be repeated to cater for different time references as previously mentioned.

app/views/admin/reports/registered_adviser/index.html.erb
```diff
@@ -0,0 +1,41 @@
+<h2>Registered Advisers</h2>
+
+<p>
+  This report shows all advisers whose FCA status is no still 'Active'.
```
margOnline: There's a typo here.

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
tomas-stefano: 👍

margOnline: I agree with @tomas-stefano that the filename method belongs in the renderer object.

tomas-stefano:
test

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
Tr4pSt3R: @margOnline the filename is an option that is merely passed to the render object by the controller though. Since the filename is first mentioned in the controller, I believe this is the best place to test it. wdyt?

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    before(:each) do
+      Timecop.freeze(date)
+    end
+
+    after(:each) do
+      Timecop.return
+    end
+
+    context 'yesterday' do
```
norm: That date is no longer yesterday, so this description no longer makes sense.

lib/ext/adviser.rb
Tr4pSt3R: @tomas-stefano the include is added outside of the `Adviser` class definitions because when it was added to the actual `Adviser` class in the **mas_rad-core** repo you advised that it should be moved out. 

Here is the changes that was proposed to the **mad_rad-core** repo (https://github.com/moneyadviceservice/mas-rad_core/pull/163/files). 

If we use a plain old ruby class to wrap the CSV functionality in what way could it open up the `Adviser` model for further addition of CSV-rendering implementations?

Is this proposal which you suggested the best way to go about this?

```ruby
class AdvisersListCsv
   def initialize(advisers)
      @advisers = advisers
   end

   def self.to_csv(_ = {})
     CSV.generate do |csv|
        csv << ['Ref. Number', 'Name', 'Firm']
 
        @advisers.each do |adviser|
          csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
        end
     end
   end
end
```

willhall88: perhaps we could avoid constraining the `.to_csv` method entirely and use a custom method instead (`generate_registered_advisers_csv`)?

tomas-stefano: With the wrapper you manipulate (or you can call extend in a gross way) *without modifying the model*. This opens ways to add more features without the *need to modify the model or touch the to_csv method*.  Imho, this respects the open/closed principle. There is a link on the wiki page on the PR review that @margOnline created that would be interested for you to read: https://maswiki.valiantyscloud.net/display/NewDev/Development+workflow
It is called "Checklist for reviewing code" that there is a small part on open closed principle.

willhall88: or are we able to use the `class_attribute` from **rad**, to extend the Adviser class in **mas_rad-core**? http://api.rubyonrails.org/classes/Class.html#method-i-class_attribute

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
Tr4pSt3R: ♻️  @norm && @willhall88 this has been fixed here: https://github.com/moneyadviceservice/rad/pull/406/commits/5e40bb7d84f3a5b24f1f6f41bfd070ff95b970e2

lib/ext/adviser.rb
Tr4pSt3R: @tomas-stefano @willhall88 

My question is... what's the best way to go about this? There has been varying expressions of opinions but there's no definite route to moving this implementation forward.

willhall88: @Tr4pSt3R @tomas-stefano I think looking at the current implementation of `.to_csv`, we've actually made it very custom to creating a specific csv file, which then constrains any further extension. IMO that perhaps means that we should not be extending the `.to_csv` method but renaming it as a new custom method.

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
Tr4pSt3R: ♻️ @norm && @willhall88 : how about this rather terse context definitions here https://github.com/moneyadviceservice/rad/pull/406/commits/2e87d92add386ddb66f7b06d4df3b1e4a04491da. Do they communicate the context of these tests well enough or is there further room for improvement?

norm: I disagree — the problem is that as a human "yesterday" has a very specific meaning. By setting another forced date in the tests, you have done nothing to remove the confusion that "yesterday" in this context is not "yesterday".

IMO you only need two tests, with very different arbitrary dates (one to prove the filename contains the date, the second to prove that it wasn't hardcoded to that date but is actually dependent on the system date at the time code is executed). Doing yesterday/today/tomorrow is cute but ultimately confusing once enough time has passed.

lib/ext/adviser.rb
tomas-stefano: My only opinion was about the use of AdvisersListCsv rather than the module one. The modules are good when it is possible to extend and it is not the case (as I explained above). The problem with just renaming is that than you have a lack of some sort of interface or the duck-typing is lost.

willhall88: resulting in something like this in the controller.. 
```
def index
  @registered_advisers = Adviser.all.page(params[:page])
  respond_to do |format|
    format.html
    format.csv { send_data Adviser.generate_registered_advisers_csv }
  end
end
```

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
```diff
@@ -8,23 +8,23 @@
       Timecop.return
     end
 
-    context 'yesterday' do
+    context 'past date' do
```
willhall88: Looking at this again now. All three tests (past, present, future) are identical and we should strip this down to just one test

willhall88: wdyt @Tr4pSt3R ?

app/controllers/admin/reports/registered_adviser_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Admin
+  module Reports
+    class RegisteredAdviserController < Admin::ApplicationController
+      respond_to :html, :csv
+
+      def index
+        @registered_advisers = Adviser.all.page(params[:page])
+        @registered_advisers_list = AdviserListCsv.new(Adviser.all)
```
tomas-stefano: Can we pass on initialize, the @registered_advisers instead? Will be less hit on the DB.

config/application.rb
```diff
@@ -1,6 +1,7 @@
 require File.expand_path('../boot', __FILE__)
 
 # Pick the frameworks you want:
+require 'csv'
```
tomas-stefano: is this require necessary? Is Rails not loading the csv? And if it is necessary, can we move the renderer initializer?

config/application.rb
```diff
@@ -9,6 +10,8 @@
 require "sprockets/railtie"
 # require "rails/test_unit/railtie"
 
+require_relative '../lib/ext/adviser_list_csv'
```
tomas-stefano: Can we move this require to the class that uses the class?

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
tomas-stefano: @Tr4pSt3R  what it is the decision on this?

Tr4pSt3R: I discussed this with @margOnline and she's happy to keep the tests where they are and as a private method in the controller where it originates from.

config/routes.rb
```diff
@@ -82,6 +82,7 @@
         end
       end
       resource :inactive_adviser, only: [:show]
+      resources :registered_adviser, only: [:index]
```
willhall88: this can just be `resource` rather than a multiple right?

Tr4pSt3R: Yeah, it breaks the tests when you change from resources to resource.

willhall88:
👍  I'm happy with this

spec/features/admin/export_advisers_list_spec.rb
```diff
@@ -0,0 +1,96 @@
+# Feature: Advisors list export
+#  As the person maintaining the RAD database.
+#  In order to get rid of unwanted advisors.
+#  I need to be able to cross reference a large list (4k) of unauthorised advisers against records in the RAD database.
+#  The list needs to be cleared up so customers do not get sent to unapproved advisers and RAD maintain it's credibility.
+#
```
margOnline: Does this comment really need to be included?

config/routes.rb
Tr4pSt3R: @willhall88 I'm using resources because the `index` action is a plural resource that will handle generic incoming requests. This is contrary to default actions such as `show`, `new`, `create`, `destroy` and ``update` which works on single resources.

margOnline:
👍

spec/features/admin/export_advisers_list_spec.rb
Tr4pSt3R: There's no fast rules about this but I like to include the direct TP story details within the feature itself so that it retains the original business requirement. 

I think they should remain to communicate the rationale behind this feature.

margOnline: I'm not sure the rationale for the feature should be included in the code, but as we have no standards, your call. We can discuss in next week's meeting re Developer Processes/Workflow.

tomas-stefano:
👍 🚢

PR #407, opened 2017-05-26
[8218] - Bugfix: Show nearest office website when viewing a firm

tomas-stefano:
## Explanation

When you're viewing a firm that has N offices, and one of the offices has a website, the office website should appear on the firm website address.

## What happened

The website was not displaying the right office website. After further investigation we understood that *if we reindex the elastic search* the problem was solved. But *if you update the office on rad admin the bug comes back*.

## Solution

Investigating deeper more we understood that on rad admin everytime that you save an office, the app sent data to elastic search without the "website" node. 

**The code that sends to elastic search all the nodes is on master and was added on mas-rad_core version 0.1.2 but the version that was on rad was 0.0.111. Updating the version solves the problem.**

You can see the code that does this now on master (0.1.2 version) here:

1) https://github.com/moneyadviceservice/mas-rad_core/blob/master/app/models/office.rb#L43
2) https://github.com/moneyadviceservice/mas-rad_core/blob/master/app/models/office.rb#L46
3) https://github.com/moneyadviceservice/mas-rad_core/blob/master/lib/mas/firm_indexer.rb#L30
4) https://github.com/moneyadviceservice/mas-rad_core/blob/master/lib/mas/firm_repository.rb#L4
5) https://github.com/moneyadviceservice/mas-rad_core/blob/master/app/serializers/office_serializer.rb#L6

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
 gem 'mailjet'
-gem 'mas-rad_core', '0.0.111'
+gem 'mas-rad_core', '0.1.2'
 gem 'oga'
 gem 'pg'
 gem 'rails_email_validator'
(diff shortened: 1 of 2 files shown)


margOnline:
👍

PR #408, opened 2017-06-28
[WIP] Rubocop update

paulanthonywilson:
There were no Rubocop complaints with the project's original configuration, but the UC (as opposed to Tomas' gist, tbh) as the template fixed Rubocop complaints for some of the project. Some of the changes are less trivial than I'd hoped.

This was supposed to be breaking ground for upcoming work on RAD, but as that's been shelved we may want to park this where it currently is. 

As Rubocop is part of the build, then we can't allow it to fail. So the new file is `rubocop-masnew.yml`. You can run against the newer rubocop with

```
rubocop -c rubocop-masnew.yml
```
Compliant directories are:

* `app/controllers`
* `app/forms`
* `app/helpers`

(The `rubocop.yml` itself as been altered to disable new checks brought in with a version update. This is no reflection on the merits of the new checks, just that the work to make the code compliant hasn't been done yet.)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,17 +1,9 @@
 Documentation:
   Enabled: false
-Style/FrozenStringLiteralComment:
-  Enabled: false
-Metrics/BlockLength:
-  Exclude:
-  - 'spec/**/*'
 Metrics/LineLength:
-  Max: 80
-  IgnoredPatterns: ['^\s*class\s']
-  Exclude:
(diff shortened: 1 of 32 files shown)


paulanthonywilson:
Lol, the build is broken because the passing Rubocop is part of the build.

paulanthonywilson:
Shutting PR down as the broken build is too much for my sensitive nature

PR #409, opened 2017-06-28
[DEV APPROVED] Use MAS standard rubocop config, and fix errors

paulanthonywilson:
**Summary**

This adds the agreed-upon standard rubocop config file that is [already used in `universal_credit`](https://github.com/moneyadviceservice/universal_credit/pull/146), and makes all changes needed for rubocop to pass.

A lot of these commits could be squashed, but we have left them quite granular to make reviewing easier.


app/controllers/self_service/firms_controller.rb
```diff
@@ -25,8 +27,11 @@ def update
 
     def available_trading_names(firm:)
       registered_trading_names = firm.trading_names.registered
-      Lookup::Subsidiary.where(fca_number: firm.fca_number).order('lower(name)').reject do |lookup_name|
-        registered_trading_names.map(&:registered_name).include? lookup_name.name
+      Lookup::Subsidiary
+        .where(fca_number: firm.fca_number)
+        .order('lower(name)').reject do |lookup_name|
+        registered_trading_names.map(&:registered_name)
```
shaley91: I think the indentation is a little messed up here.

app/controllers/admin/reports/metrics_controller.rb
```diff
@@ -1,40 +1,38 @@
...
-      private
+  private
 
-      def csv_filename(snapshot)
-        date = snapshot.created_at
-        "snapshot-#{date.year}-#{date.month}-#{date.day}.csv"
-      end
+  def csv_filename(snapshot)
+    date = snapshot.created_at
+    "snapshot-#{date.year}-#{date.month}-#{date.day}.csv"
```
Tr4pSt3R: @paulanthonywilson put these in brackets, perhaps?

Tr4pSt3R:
Apart from that small change to the method definition above every else looks good and runs well too. Let me know when it's done and I will thumb it

margOnline:
So normally I would not bother to review this because 73 files in one pr is asking alot and imo is too risky for a pr. @paulanthonywilson please would you confirm if there are any logic/code changes or does this pr only fix rubocop offenses vis a vis the new config file?

igelstorm:
@margOnline No logic or behaviour should change, it's purely fixing rubocop offences. Most changes are just syntax, and in a few cases extracting methods.

(Paul originally opened this PR but @picandocodigo and I have worked on it most recently.)

app/controllers/admin/reports/metrics_controller.rb
igelstorm: Where do you mean? I think this looks clear as it is?

app/controllers/admin/lookup/fca_import_controller.rb
```diff
@@ -17,11 +17,10 @@ def update
     if @import
       @import.commit(params[:commit])
       flash[:notice] = "Import has been #{committed}"
-      redirect_to admin_lookup_fca_import_index_path
     else
       flash[:error] = 'Could not find last import'
-      redirect_to admin_lookup_fca_import_index_path
     end
+    redirect_to admin_lookup_fca_import_index_path
```
margOnline: nice

app/controllers/admin/reports/out_of_date_firms_controller.rb
```diff
@@ -1,24 +1,32 @@
...
-  module Reports
-    class OutOfDateFirmsController < Admin::ApplicationController
-      def index
-        @firms = Firm.joins('JOIN lookup_firms ON lookup_firms.fca_number = firms.fca_number')
-                 .where(parent: nil)
-                 .where('firms.registered_name != lookup_firms.registered_name')
-                 .order('firms.fca_number ASC')
-                 .pluck('firms.fca_number', 'lookup_firms.registered_name', 'firms.registered_name', 'firms.id')
-      end
+class Admin::Reports::OutOfDateFirmsController < Admin::ApplicationController
```
margOnline: What's the rationale for namespacing the class declaration as opposed to the nested moduling previously?

app/controllers/self_service/trading_names_controller.rb
```diff
@@ -5,8 +5,7 @@ def new
     end
 
     def create
-      @firm = initialize_firm_from_lookup_trading_name(params[:lookup_id])
-      @firm.assign_attributes(firm_params)
+      @firm = build_firm_from_lookup_trading_name
```
margOnline: great method name

app/controllers/self_service/trading_names_controller.rb
```diff
@@ -32,15 +31,23 @@ def update
...
+      flash[:notice] = I18n.t('self_service.trading_name_destroy.deleted',
+                              name: trading_name.registered_name)
 
       redirect_to :back
     end
 
     private
 
+    def build_firm_from_lookup_trading_name
+      initialize_firm_from_lookup_trading_name(params[:lookup_id]).tap do |firm|
```
margOnline: nice use of tap

margOnline:
This looks good and I appreciate that most of it is syntax changes and probably updated by rubocop automatically. However, any pr this size makes me nervous. I would suggest deploying at a time when rad is not being used.

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 RSpec.describe Admin::Reports::RegisteredAdviserController, type: :controller do
   describe '#filename' do
     context 'timestamp' do
-      let(:date) { Date.new(2017, 05, 29) }
+      let(:date) { Date.new(2017, 0o5, 29) }
```
gotbadger: typo?

app/jobs/fca_import_job.rb
```diff
@@ -76,7 +83,7 @@ def parse_error(list)
 
   def notification_text(outcomes)
     if import_successful?(outcomes)
-      "The FCA data have been loaded into RAD. Visit #{admin_lookup_fca_import_index_url} to confirm that the data looks ok" # rubocop:disable all
+      "The FCA data have been loaded into RAD. Visit #{admin_lookup_fca_import_index_url} to confirm that the data looks ok"
```
gotbadger: should probably fix this rather than disabling with`rubocop:disable Metrics/LineLength`

spec/controllers/admin/reports/registered_adviser_controller_spec.rb
gotbadger: maybe just put `5` here

app/controllers/admin/reports/out_of_date_firms_controller.rb
igelstorm: Probably to keep the line length down. There are a few lines that are just under 80 chars, and if the modules were nested like before, the extra indenting would make them go over.

app/controllers/admin/reports/metrics_controller.rb
Tr4pSt3R: @igelstorm we usually put parenthesis around parameters.

```
date = snapshot.created_at(
    "snapshot-#{date.year}-#{date.month}-#{date.day}.csv"
)
```

igelstorm: I see what you mean. That's not how this code will be interpreted though: the string isn't an argument to `created_at`, it's a separate statement (an implicit `return`).

It's first assigning `date = snapshot.created_at`, and then returning a string with some attributes of `date` interpolated.

igelstorm:
@margOnline Is there a particular time you would suggest to deploy it? I'm inclined to think, since it'll be tested on staging first, and rollbacks are very quick to do on Heroku in case there's a problem, this should still be quite safe even though it's big.

app/controllers/admin/reports/metrics_controller.rb
Tr4pSt3R: Oh wow! That is so deceptive. In that case do you mind breaking the two statements with a line so that it doesn't look like `"snapshot-#{date.year}-#{date.month}-#{date.day}.csv"` is an argument for the `created_at` method?

#option 1: 
```
date = snapshot.created_at
return "snapshot-#{date.year}-#{date.month}-#{date.day}.csv"
```

#option 2: 
```
date = snapshot.created_at

"snapshot-#{date.year}-#{date.month}-#{date.day}.csv"
```

norm: Since it's already caused confusion, perhaps that should be made explicit — `return "snapshot..."`?

willhall88: I'm happy with either of the two options @Tr4pSt3R has suggested. I think it does need to be explicit that it isn't an argument as that was my first assumption.

Perhaps we could create an alias to the `created_at` method 
`date = snapshot.date_created`

margOnline: I prefer the blank line between the 2 statements, option 2. Its more in line with standards on other MAS repos.

I was not confused by the initial definition - a string was being returned and the snapshot date was being used to make up the string. Also the method name `csv_filename` indicates that probably a string is going to be returned.

margOnline:
@igelstorm late afternoon/evening. One of the BAs should be able to say when its not being used.

app/jobs/fca_import_job.rb
willhall88: You should be able to break this up onto multiple lines without breaking anything I think by using `\`

https://github.com/moneyadviceservice/development-standards/tree/master/coding-styles/ruby#2.4

app/controllers/admin/reports/metrics_controller.rb
willhall88: @margOnline you're right, looking at the full method block it is more clear. My first assumption and comment was based solely on looking at `.created_at` rather than the full method. 

I agree with option 2 being in line with MAS code.

app/jobs/fca_import_job.rb
willhall88: I should make clear that the above link is for reference and the development-standards repo is still an unreviewed WIP project, so don't take everything written there as gospel. yet ;)

app/controllers/principals_controller.rb
```diff
@@ -24,29 +26,37 @@ def new
...
-      user = User.new(@form.user_params)
-      user.build_principal(@form.principal_params)
-      user.save!
+    form = NewPrincipalForm.new(new_principal_form_params)
 
-      Stats.increment('radsignup.principal.created')
-
-      Identification.contact(user.principal).deliver_later
-      redirect_to user.principal
+    if form.valid?
```
willhall88: I'm surprised rubocop didn't suggest a guard clause instead of this if statement.

app/jobs/fca_confirmation_job.rb
```diff
@@ -1,21 +1,25 @@
...
+
+  unique_args = ->(args) { Digest::SHA1.hexdigest(args.first.to_s) }
   sidekiq_options unique:      :until_executed,
-                  unique_args: ->(args) { Digest::SHA1.hexdigest(args.first.to_s) }
+                  unique_args: unique_args
   queue_as :default
 
   def perform(import_job_id)
-    if import = FcaImport.find_by_id(import_job_id) # rubocop:disable all
+    # rubocop:disable all
```
willhall88: what are the failing cops in this block that requires all of rubocop to be disabled?

app/controllers/admin/reports/metrics_controller.rb
gotbadger: This is probably the ideal place to use the `.strftime` method since it specifically is made to address this kind of problem. However since the main point of this PR is to fix rubocop errors rather than a lot of refactoring perhaps the line break option is more sensible.

willhall88:
extra thumbs for moving sql queries into private methods. 
Most of this looks good. 
I'd agree that we should be careful about when we deploy such a large PR. We should also be considerate of the QA's to test this as it'll need a thorough review of the whole app.

app/jobs/fca_confirmation_job.rb
igelstorm: @willhall88 There were a few, and we were reluctant to refactor this too much as we weren't sure how thorough the test coverage is for this class. A few of them are fairly trivial though, so I've pushed some more fixes. Now only one cop is explicitly disabled (which I don't think we can fix without potentially altering the behaviour).

igelstorm:
@margOnline @willhall88 I guess the risk of deploying late in the day is that fewer people will be around to notice or fix any problems, meaning they take longer to be discovered or fixed - so if there is a problem, although the user impact at any given moment will be less, the total impact over time might be greater. 

If anything is at risk of getting missed by a thorough QA, it’ll presumably be very subtle things that won’t be obvious straight away (maybe only after hearing from users), so perhaps the least risky time is actually early in the day, when there’s a bit of load and customer service staff around to notice problems? If issues do appear, a rollback in Heroku would be very quick.

paulanthonywilson:
This never got merged, and it is the thing that addresses the Rubocop vulnerability that was not dealt with in https://github.com/moneyadviceservice/rad/pull/418

Suggest we have another round of reviews.

Gemfile
```diff
@@ -32,31 +33,30 @@ gem 'mas-rad_core', '0.1.3'
...
   gem 'binding_of_caller'
   gem 'factory_girl_rails'
   gem 'faker'
-  gem 'pry-rails'
+  gem 'ffaker'
   gem 'pry-byebug'
+  gem 'pry-rails'
   gem 'rspec-rails'
-  gem 'rubocop', '0.31.0'
+  gem 'rubocop', '0.49.0'
```
Tr4pSt3R: Yes, @paulanthonywilson this PR fixes the vulnerability that was highlighted in the old version `0.31.0`.

PR #415, opened 2017-12-12
[DEV APPROVED] 8604 update fca import

tomas-stefano:
TP: https://moneyadviceservice.tpondemand.com/entity/8604

## Context

The FCA Data Extract team recently made some changes to the files that they will be sending in the future. As of January 2018 we will no longer receive the firms file which, the FCA import automation script is hard-coded to pick up. 

Basically instead of picking the firms file we should be choosing the firms master file.
The firms master file came in different encoding format than the older file but the rest are almost the same.

## References

I opened a PR about documentation of FCA on https://github.com/moneyadviceservice/technical-docs/pull/2/

### DIFF
--- a/fca_import_scripts/lib/ext_to_sql.rb
+++ b/fca_import_scripts/lib/ext_to_sql.rb
@@ -48,7 +48,7 @@ class ExtToSql
     case row[COLUMNS::HEADER_NAME]
     when 'Individual Details'
       :adviser
-    when 'Firm Authorisation'
+    when 'Firm Authorisation', 'Firms Master List'
       :firm
     when 'Alternative Firm Name'
       :subsidiary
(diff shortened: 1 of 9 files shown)


fca_import_scripts/lib/ext_to_sql.rb
```diff
@@ -95,7 +95,7 @@ def repairs
   end
 
   def repair_line(line)
-    line = line.strip
+    line = line.strip.mb_chars.tidy_bytes
```
margOnline: nice

margOnline:
Other than cleaning up the commit history, lgtm 👍

tomas-stefano:
Thanks. What exactly need cleaning up on the commits?

margOnline:
I was referring to the phantom js version change, but on second look perhaps it should be a separate commit.

tomas-stefano:
Ok, I will leave as it is, since it was an isolated issue.

willhall88:
👍

PR #416, opened 2018-01-08
[DEV APPROVED] 8625 Adds qualifications and accreditations to report

KateWilkinson:
[TP Story](https://moneyadviceservice.tpondemand.com/entity/8625)

Adds Qualifications and Accreditations to the Registered Advisers report, so that against each advisor there is a new column for each Accreditation and qualification with a Y/N in each field for whether or not the qualification is held.

**EDIT**

Also adds an `AdviserCsvLookup`, which loads all the necessary data from the db into memory before we transform it into the correct state for the CSV. `ActiveRecord`'s built in-caching and preloading was unable to give us anything close to the necessary performance - this speeds up the generation of the CSV file (from about 35s down to about 2s).

### DIFF
--- a/lib/adviser_csv_lookup.rb
+++ b/lib/adviser_csv_lookup.rb
@@ -1,31 +1,33 @@
 class AdviserCsvLookup
-  attr_reader :qualifications_lookup, :accreditations_lookup, :firms_lookup
-
   def initialize(advisers)
     @advisers = advisers
   end
 
-  def build!
-    @qualifications_lookup = build_adviser_qualifications
-    @accreditations_lookup = build_adviser_accreditations
-    @firms_lookup          = build_adviser_firms
-  end
-
   def certification_list
-    qualifications.values + accreditations.values
+    all_qualifications + all_accreditations
   end
 
-  def each(&block)
+  def each
     @advisers.each do |adviser|
-      adviser_qualifications = qualifications_lookup[adviser.id] || []
-      adviser_accreditations = accreditations_lookup[adviser.id] || []
-      block.call(
-        adviser: adviser,
-        firm_name: firms_lookup[adviser.firm_id],
-        qualifications: adviser_qualifications,
-        accreditations: adviser_accreditations
-      )
+      adviser_qualifications = qualifications_lookup.for(adviser.id)
+      adviser_accreditations = accreditations_lookup.for(adviser.id)
+      yield(adviser: adviser,
+            firm_name: firms_lookup[adviser.firm_id],
+            qualifications: adviser_qualifications,
+            accreditations: adviser_accreditations)
     end
+  end
+
+  def qualifications_lookup
+    @qualifications_lookup ||= AdviserQualificationsLookup.new(adviser_ids)
+  end
+
+  def accreditations_lookup
+    @accreditations_lookup ||= AdviserAccreditationsLookup.new(adviser_ids)
+  end
+
+  def firms_lookup
+    @firms_lookup ||= Firm.pluck(:id, :registered_name).to_h
   end
 
   private
@@ -34,41 +36,11 @@ class AdviserCsvLookup
     @advisers.map(&:id)
   end
 
-  def build_adviser_qualifications
-    certs = certification_query('advisers_qualifications', 'qualification')
-    aggregate_certificate_names(certs, type: 'qualification')
+  def all_qualifications
+    qualifications_lookup.all_certifications
   end
 
-  def build_adviser_accreditations
-    certs = certification_query('accreditations_advisers', 'accreditation')
-    aggregate_certificate_names(certs, type: 'accreditation')
-  end
-
-  def build_adviser_firms
-    Firm.pluck(:id, :registered_name).to_h
-  end
-
-  def certification_query(table, key)
-    ActiveRecord::Base.connection.execute("
-      SELECT adviser_id, #{key}_id
-      FROM #{table}
-      WHERE adviser_id IN (#{adviser_ids.join(',')})
-    ").to_a
-  end
-
-  def qualifications
-    @qualifications ||= Qualification.pluck(:id, :name).to_h
-  end
-
-  def accreditations
-    @accreditations ||= Accreditation.pluck(:id, :name).to_h
-  end
-
-  def aggregate_certificate_names(adviser_certifications, type:)
-    adviser_certifications.each_with_object({}) do |certification, h|
-      adviser_id = certification['adviser_id'].to_i
-      h[adviser_id] ||= []
-      h[adviser_id] << send(type.pluralize)[certification["#{type}_id"].to_i]
-    end
+  def all_accreditations
+    accreditations_lookup.all_certifications
   end
 end
--- a/lib/adviser_csv_lookup.rb
+++ b/lib/adviser_csv_lookup.rb
@@ -1,0 +1,74 @@
+class AdviserCsvLookup
+  attr_reader :qualifications_lookup, :accreditations_lookup, :firms_lookup
+
+  def initialize(advisers)
+    @advisers = advisers
+  end
+
+  def build!
+    @qualifications_lookup = build_adviser_qualifications
+    @accreditations_lookup = build_adviser_accreditations
+    @firms_lookup          = build_adviser_firms
+  end
+
+  def certification_list
+    qualifications.values + accreditations.values
+  end
+
+  def each(&block)
+    @advisers.each do |adviser|
+      adviser_qualifications = qualifications_lookup[adviser.id] || []
+      adviser_accreditations = accreditations_lookup[adviser.id] || []
+      block.call(
+        adviser: adviser,
+        firm_name: firms_lookup[adviser.firm_id],
+        qualifications: adviser_qualifications,
+        accreditations: adviser_accreditations
+      )
+    end
+  end
+
+  private
+
+  def adviser_ids
+    @advisers.map(&:id)
+  end
+
+  def build_adviser_qualifications
+    certs = certification_query('advisers_qualifications', 'qualification')
+    aggregate_certificate_names(certs, type: 'qualification')
+  end
+
+  def build_adviser_accreditations
+    certs = certification_query('accreditations_advisers', 'accreditation')
+    aggregate_certificate_names(certs, type: 'accreditation')
+  end
+
+  def build_adviser_firms
+    Firm.pluck(:id, :registered_name).to_h
+  end
+
+  def certification_query(table, key)
+    ActiveRecord::Base.connection.execute("
+      SELECT adviser_id, #{key}_id
+      FROM #{table}
+      WHERE adviser_id IN (#{adviser_ids.join(',')})
+    ").to_a
+  end
+
+  def qualifications
+    @qualifications ||= Qualification.pluck(:id, :name).to_h
+  end
+
+  def accreditations
+    @accreditations ||= Accreditation.pluck(:id, :name).to_h
+  end
+
+  def aggregate_certificate_names(adviser_certifications, type:)
+    adviser_certifications.each_with_object({}) do |certification, h|
+      adviser_id = certification['adviser_id'].to_i
+      h[adviser_id] ||= []
+      h[adviser_id] << send(type.pluralize)[certification["#{type}_id"].to_i]
+    end
+  end
+end
--- a/lib/adviser_list_csv.rb
+++ b/lib/adviser_list_csv.rb
@@ -4,7 +4,6 @@ class AdviserListCsv
   end
 
   def to_csv(_ = {})
-    @lookup.build!
     CSV.generate do |csv|
       csv << headings
       @lookup.each do |adviser_data|
--- a/lib/adviser_list_csv.rb
+++ b/lib/adviser_list_csv.rb
@@ -1,13 +1,14 @@
 class AdviserListCsv
   def initialize(advisers)
-    @advisers = advisers
+    @lookup = AdviserCsvLookup.new(advisers)
   end
 
   def to_csv(_ = {})
+    @lookup.build!
     CSV.generate do |csv|
       csv << headings
-      @advisers.each do |adviser|
-        csv << row(adviser)
+      @lookup.each do |adviser_data|
+        csv << row(adviser_data)
       end
     end
   end
@@ -18,25 +19,23 @@ class AdviserListCsv
     ['Ref. Number', 'Name', 'Firm'] + all_certifications
   end
 
-  def row(adviser)
+  def row(adviser:, firm_name:, qualifications:, accreditations:)
     [
       adviser.reference_number,
       adviser.name,
-      adviser.firm.registered_name
-    ] + certified(adviser)
+      firm_name
+    ] + certified(qualifications, accreditations)
   end
 
-  def all_certifications
-    @certifications ||= Accreditation.pluck(:name) + Qualification.pluck(:name)
-  end
+  def certified(qualifications, accreditations)
+    adviser_certs = qualifications + accreditations
 
-  def certified(adviser)
     all_certifications.map do |certification|
-      certifications_for(adviser).include?(certification) ? 'Y' : 'N'
+      adviser_certs.include?(certification) ? 'Y' : 'N'
     end
   end
 
-  def certifications_for(adviser)
-    adviser.accreditations.pluck(:name) + adviser.qualifications.pluck(:name)
+  def all_certifications
+    @all_certifications ||= @lookup.certification_list
   end
 end
--- a/lib/adviser_list_csv.rb
+++ b/lib/adviser_list_csv.rb
@@ -5,13 +5,38 @@ class AdviserListCsv
 
   def to_csv(_ = {})
     CSV.generate do |csv|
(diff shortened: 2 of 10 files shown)


lib/adviser_list_csv.rb
```diff
@@ -5,11 +5,33 @@ def initialize(advisers)
...
   end
+
+  private
+
+  def headings
+    ['Ref. Number', 'Name', 'Firm'] + all_certifications
+  end
+
+  def all_certifications
+    @all_certifications ||= Accreditation.all.pluck(:name) + Qualification.all.pluck(:name)
```
tomas-stefano: Minor detail. Can you remove the all from both calls? Just `Accreditation.pluck(:name)` and `Qualification.pluck(:name)` will work fine.

lib/adviser_list_csv.rb
```diff
@@ -5,11 +5,33 @@ def initialize(advisers)
...
+  end
+
+  def certified(adviser)
+    adviser_certifications = adviser.accreditations.pluck(:name) + adviser.qualifications.pluck(:name)
+    all_certifications.map { |name| adviser_certifications.include?(name) ? 'Y' : 'N' }
```
tomas-stefano: Another minor. Can we change the map variable from `name` to `certification`? Wdyt?

spec/lib/adviser_list_csv_spec.rb
```diff
@@ -1,7 +1,34 @@
...
+    let(:adviser1) { FactoryGirl.create(:adviser, qualifications: [qualifications.first], accreditations: [accreditations.first]) }
+    let(:adviser2) { FactoryGirl.create(:adviser, qualifications: [qualifications.last], accreditations: [accreditations.last]) }
+    let(:subject)  { described_class.new(advisers) }
+
     it 'responds to csv conversion' do
-      expect(described_class.new []).to respond_to(:to_csv)
+      expect(subject).to respond_to(:to_csv)
+    end
+
+    it 'produces correct data' do
```
tomas-stefano: My feeling is that this spec is become a little bigger. Can we break into smaller specs? One for the header and one for the data? OR can we rename this to something related to: 'returns qualifications from each advisor'?

KateWilkinson: I've made a couple of changes to the spec @tomas-stefano, please could you have another look and see if this is any clearer?

lib/adviser_list_csv.rb
KateWilkinson: 👍  sounds good, will change now

lib/adviser_list_csv.rb
KateWilkinson: 👍

margOnline: please address the line length

lib/adviser_list_csv.rb
```diff
@@ -5,11 +5,33 @@ def initialize(advisers)
...
+  def headings
+    ['Ref. Number', 'Name', 'Firm'] + all_certifications
+  end
+
+  def all_certifications
+    @all_certifications ||= Accreditation.all.pluck(:name) + Qualification.all.pluck(:name)
+  end
+
+  def certified(adviser)
+    adviser_certifications = adviser.accreditations.pluck(:name) + adviser.qualifications.pluck(:name)
```
margOnline: please address the line length for this line and the one below it

spec/lib/adviser_list_csv_spec.rb
```diff
@@ -1,7 +1,47 @@
 RSpec.describe AdviserListCsv do
   describe '#to_csv' do
+    let(:qualifications)     { 3.times.map { FactoryGirl.create(:qualification) } }
+    let(:accreditations)     { 3.times.map { FactoryGirl.create(:accreditation) } }
+    let(:all_certifications) { Accreditation.pluck(:name) + Qualification.pluck(:name) }
+
+    let(:advisers) { [adviser1, adviser2] }
+    let(:adviser1) { FactoryGirl.create(:adviser, qualifications: [qualifications.first], accreditations: [accreditations.first]) }
+    let(:adviser2) { FactoryGirl.create(:adviser, qualifications: [qualifications.last], accreditations: [accreditations.last]) }
```
margOnline: please address the line length

spec/lib/adviser_list_csv_spec.rb
tomas-stefano: Sorry to be picky but can you rename the specs to something like "returns csv headings" instead of using "has correct headings"? And the same for advisers data specs. I am not sure about others, but "has correct ..." on specs sounds redundant to me.

lib/adviser_list_csv.rb
KateWilkinson: Hi @margOnline. I'm just going through and sorting out the line lengths now.
Metrics/LineLength seems to be set to 120 in the RAD rubocop file, but in the example from the PR Approval doc it's 80. Shall I update this in RAD or would you prefer I just run against the example rubocop file instead?

spec/lib/adviser_list_csv_spec.rb
KateWilkinson: No worries, can do :)

lib/adviser_list_csv.rb
```diff
@@ -5,11 +5,38 @@ def initialize(advisers)
 
   def to_csv(_ = {})
     CSV.generate do |csv|
-      csv << ['Ref. Number', 'Name', 'Firm']
-
+      csv << headings
       @advisers.each do |adviser|
-        csv << [adviser.reference_number, adviser.name, adviser.firm.registered_name]
+        csv << row(adviser)
```
margOnline: nice

lib/adviser_list_csv.rb
margOnline: Good question. I have a feeling that we haven't updated it in the RAD repo because it would throw too many errors. 

We are try to update to the new standards with every file we touch. As per the previous engagement with Cultivate, we ask that you all do the same. Appreciate that its tedious but its all part of the job. Perhaps this is something we can discuss when you are in London later this week?

If you dont mind updating the RAD rubocop file and running it just to see what it throws, that would be great.

Tr4pSt3R:
Hi @KateWilkinson it's a shame to hear that the first deploy didn't go as planned. I was wondering, for this PR, whether you have pulled down the RAD production DB to test the new functionality locally to ensure that it doesn't happen in the next deploy?

KateWilkinson:
Hi @Tr4pSt3R, good point. Had a look and the number of Advisers in staging is pretty close to the number in production, so it might be most straightforward to get this up onto staging again and test it out there rather than get a dump of production data? I'm not sure how quick/straightforward it is to get the prod data down from RAD?

lib/adviser_list_csv.rb
```diff
@@ -5,11 +5,30 @@ def initialize(advisers)
...
       end
     end
   end
+
+  private
+
+  def headings
+    ['Ref. Number', 'Name', 'Firm', 'Accreditations', 'Qualifications']
```
willhall88: Could we use `%w` rather than a literal array syntax here please? 
https://github.com/moneyadviceservice/development-standards/tree/master/coding-styles/ruby#11.3

Should this array also be frozen?

KateWilkinson: 👍

Tr4pSt3R:
@KateWilkinson there's a documentation here on how to mirror your local database with production database. Although, I would advise, that the _prod_ DB dump and the _prod-populated_ DB should be purged for security reasons. 

https://maswiki.valiantyscloud.net/pages/viewpage.action?pageId=63635527

lib/adviser_list_csv.rb
KateWilkinson: @willhall88 is `%w` still preferred when I need to escape the space in `Ref. Number`?

willhall88: ah, my bad. Missed that one. ok, happy as it currently is. But it still looks to me like it should be a frozen array?

KateWilkinson: No worries, just thought I'd check what the preference is! Yep I'll change that 👍

spec/lib/adviser_list_csv_spec.rb
```diff
@@ -1,13 +1,68 @@
...
+          'N', 'N', 'Y', 'N', 'N', 'Y'
+        ])
+      end
     end
-  end
 
-  context 'Office' do
-    it 'should not respond to csv conversion' do
-      expect(Office).to_not respond_to(:to_csv)
+    context 'Office' do
```
Tr4pSt3R: This should be a `describe` rather than a `context`. The question is.. does `Office` ever respond to the `#csv` method? If it does then we need to add that context. If it doesn't then a `describe` which ascertains that neither the `Office` or any instance of the `Office` class responds to `csv`. 

I hope it makes sense. If doesn't we can have a quick call over Slack to discuss this.

gotbadger: `describe` is used for methods `to_csv` is the method and in this case we are testing it with `Office` so feel like what we are doing here is fine?

Generally I feel this test is kinda strange (it was from the original code) and I would rather we remove it totally.

spec/lib/adviser_list_csv_spec.rb
```diff
@@ -1,13 +1,68 @@
...
+        qualifications: [qualifications.last],
+        accreditations: [accreditations.last]
+      }
+    end
+
+    let(:subject)  { described_class.new(advisers) }
+
     it 'responds to csv conversion' do
-      expect(described_class.new []).to respond_to(:to_csv)
+      expect(subject).to respond_to(:to_csv)
```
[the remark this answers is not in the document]
gotbadger: if you think this helps then yeah we can do this

spec/lib/adviser_list_csv_spec.rb
tomas-stefano: @Tr4pSt3R perhaps I miss understood. This lines were already there and introduced here: https://github.com/moneyadviceservice/rad/commit/a50e1badca4e593f56c2008488423454639074bf#diff-59aaa67083461a3d95ed7e719c81ca06R8

spec/lib/adviser_list_csv_spec.rb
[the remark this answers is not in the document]
tomas-stefano: IMHO this test should not exist because it is already testing the `to_csv` on all the tests below. It makes redundant.

spec/lib/adviser_list_csv_spec.rb
Tr4pSt3R: @tomas-stefano the lines were there but I'm saying it should look something like this

```ruby
   1 RSpec.describe AdviserListCsv do
   2   describe '#to_csv' do
   3     it 'responds to csv conversion' do
   4       expect(described_class.new []).to respond_to(:to_csv)
   5     end
   6
   7     context 'Office' do
   8       it 'should not respond to csv conversion' do
   9         expect(Office).to_not respond_to(:to_csv)
  10       end
  11     end
  12   end
  13 end
```
When you see it as written above then you'd see why having that context above is truly redundant.  This test for `Office` responding to the `to_csv` method should be scrapped.

Rather than, something like this...
```ruby
   1 RSpec.describe AdviserListCsv do
   2   describe '#to_csv' do
   3     it 'responds to csv conversion' do
   4       expect(described_class.new []).to respond_to(:to_csv)
   5     end
   6   end
   7
   8   context 'Office' do
   9     it 'should not respond to csv conversion' do
  10       expect(Office).to_not respond_to(:to_csv)
  11     end
  12   end
  13 end
```

At any rate this test that the Office class does not respond to `to_csv` is unnecessary.

tomas-stefano: This two specs are unnecessary. Usually we use respond_to spec when we want to force some kind of interface, which is not the case.

gotbadger: removed `respond_to` tests 👍

Tr4pSt3R:
@tomas-stefano would you like to re-thumb this PR and label it with [DEV APPROVED], if you are happy with it?

lib/adviser_csv_lookup.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def aggregate_certificate_names(adviser_certifications, type:)
+    adviser_certifications.each_with_object({}) do |certification, h|
+      adviser_id = certification['adviser_id'].to_i
+      h[adviser_id] ||= []
+      h[adviser_id] << send(type.pluralize)[certification["#{type}_id"].to_i]
+    end
+  end
+end
```
tomas-stefano: I will be honest that was kind of difficult to get my head around it. So I have questions and considerations about this class:

1) Is it not possible to make any sort of inner join from Adviser with accreditations and qualifications? This class sounds to me something wrong with the table relationship or with the mapping of the database, but I can be wrong. I am asking because then almost everything of this class could be removed if the answer is yes.
2) Do we need raw SQL to run a simple select query? Can Active Record be used?
3) There is the build! method with is called in the other class before the each method. This sound too much couple to do in other class. Can we not put inside the each method?
4) This class does not respect the single responsibility principle. There is a lot of going on on building the accreditation and qualification for the adviser plus getting the certification list plus aggregating the certificated names. I still need to think what could be a desired solution, but perhaps has one CertificationList class and this one.
5) I still didn't figured it out yet, but since we have a lot of private methods, perhaps there is some tests missing covering the behaviour. Like the aggregated method, where is it being tested?

Sorry, don't take personally, I understand the performance improvements, but I think we need to re look at this file. I would like to hear @margOnline opinions as well.

scott-ad-riley: Not a problem - answers below:

1. So `ActiveRecord`'s own cache was not fast enough for this to run in reasonable time. This meant that when we fetched an advisor and then attempted to fetch their qualifications/accreditations and their names, going across this many to many relation meant essentially a pretty bad bad n + 1.

We did attempt to write this as an inner join but, because of the join table in the middle, and the fact we wanted to actually output a list of all qualifications/accreditations filtered against the ones on an advisor the raw sql would have been incredibly complex and hard to follow.

2. The query is not simple because we're querying off of a join table (i.e. not a model). It would have been nice to be able to do:
```ruby
AdvisorQualifications.where(adviser_id: adviser_ids)
# and
AdvisorAccreditations.where(advisor_id: adviser_ids)
```
However this would involve creation of a new class, and essentially moving complication elsewhere.

3. Yep, I can move this 👍 

4. I can also have a look at splitting this apart into smaller classes too.

5. The classes is essentially tested on input/output. We're not testing the `each` method here - we made the lookup hashes for adviser qualifications, accreditations and firms available via `attr_readers` and test their specific structure and contents. This is the test that exercises the aggregate method. The private methods are the results of reducing duplication and refactoring, not adding additional untested behaviour.

Not a problem - we definitely would have liked to have stuck with `ActiveRecord`, but it was very clear from the outset that it wasn't able to deal with the many to many join table efficiently.

tomas-stefano:
Without doubt it is more easier to follow, but there is still duplication going on between AdviserAccreditationsLookup and AdviserQualificationsLookup.

scott-ad-riley:
Ok - pushed up a fix for the duplication (extracting a `AdviserCertificationLookup` module)

tomas-stefano:
I approved but it please tidy up the commits.

spec/lib/adviser_csv_lookup_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    {
+      qualifications: [qualifications.last],
+      accreditations: [accreditations.last]
+    }
+  end
+
+  let(:subject) { described_class.new(advisers) }
+
+  describe '#each' do
+    let(:perform) { subject.each {} }
```
margOnline: I'm not sure `perform` is the best name for this variable.

spec/lib/adviser_qualifications_lookup_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      qualifications: [qualifications.first, qualifications.last]
+    }
+  end
+
+  let(:subject) { described_class.new([adviser.id]) }
+
+  describe '#for' do
+    it 'returns the qualifications for an advisor' do
+      expect(subject.for(adviser.id)).to eq(
+        [qualifications.first.name, qualifications.last.name]
```
margOnline: It would be better for the `eq` clause to contain the actual expected values

spec/lib/adviser_csv_lookup_spec.rb
KateWilkinson: Hi @margOnline, happy to change this, but not sure what a better alternative would be here. Do you have any ideas?

spec/lib/adviser_qualifications_lookup_spec.rb
KateWilkinson: 👍 Ok, will change this.

KateWilkinson: @margOnline because the qualification & accreditation factories generate names based on ids, eg. `Qualification 1` it's tricky to test those values reliably when the ids will change. I could get around it by passing a specific name in when we create the qualifications/accreditations at the top of the specs - would that be preferable?

PR #418, opened 2018-01-11
[DEV APPROVED] Upgrade Nokogiri and other packages to remove security vulnerabilities

paulanthonywilson:
Github was reporting a number of vulnerabilities:

voila_capture 2018-01-09_10-12-55_am

Most of the update was done by @osharpe. There are some additional changes to turn the build back to green. These are currently separated into different explanatory commits. They can be squashed, but I would prefer to use Github's "squash and merge" to preserve the history.

Note that:

* Rubcop is pinned to an earlier version. It does not seem appropriate to update and fix the subsequent violations in this PR.
* There are numerous warnings associated with `FactoryGirl`. Fixing those will require a change to [mas-rad_core](https://github.com/moneyadviceservice/mas-rad_core).
* A rebase against `master` was needed to incorporate https://github.com/moneyadviceservice/rad/pull/415


spec/features/admin/move_advisers_spec.rb
```diff
@@ -51,13 +51,10 @@ def and_i_want_to_move_adviser(adviser)
 
   def and_i_want_to_move_to_firm(firm)
     expect(choose_firm_page).to be_displayed
-    expect(choose_firm_page.hidden.advisers[0].value).to eq(adviser.id.to_s)
```
willhall88: I've not looked at the full scenario of what this should be doing, but rather than just deleting these tests, should we be looking to cover this in another way? 
I.e. checking that we cannot see certain advisers on the page?

paulanthonywilson: If there were much value in the checks, I'd tend to agree.

But this is an end-to-end spec about moving and adviser from one firm to another, which will fail if those hidden inputs are not present, the checks are redundant.

On balance I can't see it being worth the overhead of using something other than Capybara for checking the html in this one case. It will add to the overhead of understanding the code.

paulanthonywilson: @willhall88 If you did want to retain these checks, what do you reckon we should use? 

(I'm thinking regex on the html.)

willhall88: ok, if you've checked and are sure the tests will still fail and those expectations are redundant then I am happy that they can go.

paulanthonywilson: ✌️ Yes, totally fails if we mess with the production code so that the hidden advisers and firms aren't present.

Gemfile
```diff
@@ -53,7 +54,7 @@ group :test, :development do
   gem 'pry-rails'
   gem 'pry-byebug'
   gem 'rspec-rails'
-  gem 'rubocop'
+  gem 'rubocop', '0.31.0'
```
Tr4pSt3R: @paulanthonywilson according to the PR title this PR is to upgrade Nokogiri `Upgrade Nokogiri and other packages to remove security vulnerabilities #418`. Am I right in assuming that this involves everything reported by GitHub on this repo as you've described in the description for this PR?

- [x] Rails > 4.2.5.2
- [x] JQuery-Rails > 4.0.3
- [x] Nokogiri > 1.6.7.2
- [ ] Rubocop > 0.31.0
- [x] RubyZip > 1.2.0

_Rubocop_ has been pinned to `0.31.0` in `Gemfile` as `gem 'rubocop', '0.31.0'`. Shouldn't this version of Rubocop be upgraded too?

paulanthonywilson: @Tr4pSt3R  I totally forgot that Rubocop was one of the vulnerabilities, so yes - good point.

Nether-the-less, I think it would make sense to upgrade Rubocop in another PR. Upgrading introduces a lot of rules and subsequent violations, so I'd rather have a go at fixing those in its own PR.

(Also the vulnerability is slight, and rubocop is only included in dev and test.)

Tr4pSt3R: @paulanthonywilson yeah sure. We can make a separate PR to deal with Rubocop

PR #419, opened 2018-01-18
Upgrade mas-rad_core from Github master.

paulanthonywilson:
The current gem version is broken in production as it uses the no longer available Redis.connect function in the geocoder initialiser.

The changes to mas-rad_core also removes many deprecation warnings which were appearing while running the specs.

Related to https://moneyadviceservice.tpondemand.com/entity/8699-rad-upgrade-rails-and-node-packages

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,8 +29,7 @@ gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
 gem 'mailjet'
-gem 'mas-rad_core', git: 'https://github.com/moneyadviceservice/mas-rad_core.git',
-                    tag: 'v0.1.4'
+gem 'mas-rad_core', '0.1.4'
 gem 'oga'
 gem 'pg'
 gem 'rails_email_validator'
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,8 @@ gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
 gem 'mailjet'
-gem 'mas-rad_core', '0.1.3'
+gem 'mas-rad_core', git: 'https://github.com/moneyadviceservice/mas-rad_core.git',
+                    tag: 'v0.1.4'
 gem 'oga'
 gem 'pg'
 gem 'rails_email_validator'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,6 +15,23 @@ GIT
       rails (>= 3.2)
       sass-rails (~> 4.0.0)
 
+GIT
+  remote: https://github.com/moneyadviceservice/mas-rad_core.git
+  revision: 1dc54b8829a8a8f260bc45bf5dc1ec4405921f07
+  tag: v0.1.4
+  specs:
+    mas-rad_core (0.1.4)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -29,7 +29,8 @@ gem 'jquery-rails'
 gem 'kaminari'
 gem 'letter_opener', group: :development
 gem 'mailjet'
-gem 'mas-rad_core', '0.1.3'
+gem 'mas-rad_core', git: 'https://github.com/moneyadviceservice/mas-rad_core.git',
```
margOnline: My instinct says we should use rubygems for the source of our gems. Can we not fix the version on rubygems?

tomas-stefano: I agree with the Rubygems process. I am afraid that using git will be only disguise a larger problem that should be addressed. 

**The actual problem**: the code is "shared" between the apps but only parts of it. There are parts that belongs to rad consumer and parts that belongs to rad admin. Everything in one gem.

We should be clearing the tech debt as we go along, gradually. So saying that I think the best approach is to we (all backend devs) talk and see a better appproach to this *big* tech debt.

Shall we organize a call to talk about this subject? Wdyt?

paulanthonywilson: I don't know why we haven't noticed this before, but publishing a gem that's of no use to anyone else seems like an abuse of open source infrastructure to me. Also - what's the point? It just adds an unnecessary step to the process and is inconsistent with the way that Dough is included in this project.

> the code is "shared" between the apps but only parts of it. There are parts that belongs to rad consumer and parts that belongs to rad admin

Is that true? It could be; I'm not that familiar with much of RAD. If so then bringing in RAD only code to RAD, and the same to consumer would be A Good Thing(tm).

I've thought for a while that bringing the two sides into the same repository and deployment, in some way that separated responsibilities, would be better, but that's a pretty big job.

tomas-stefano: Yes it will be a big job. I agree.
But, I still think it is best we continue to publish to Rubygems.
Because there is no advantage besides removing one command line `gem push`. IMHO, it is more worth it to create spikes to investigate the current state of mas rad core and which parts can be taken/moved out, **gradually** than remove one part of the process that runs in a command line.
And other thing, I don't know the plans of the sysadmins of moving rad out from Heroku and use Mas infrastructure to build the gem and publish in the private rubygems server and also moving rad to our infrasctructure.  
Sorry, I don't want to be picky on this subject, just state my opinions. Wdyt?

paulanthonywilson: Ok, I don't think it makes any sense but I've changed the Gemfile to point to `mas-rad_core` `0.1.4` on ruby gems. Can someone with permission push that gem please?

tomas-stefano: What do you think does not makes sense?

About the permission, the development team account has permission (on keepass) to add yourself as contributor to the project. A document that could be useful on this case is this one [here](https://maswiki.valiantyscloud.net/display/RRAD/Build+and+release+mas-rad_core).

paulanthonywilson: I've pushed the fixed version of `mas-rad_core` to RubyGens as MAS

tomas-stefano:
👍

PR #421, opened 2018-01-26
Update Dockerfile & Jenkins test script for Ubuntu LTS

osharpe:
The Jenkins Docker image utilised Ubuntu 17.04, which has become end of life. This PR rolls back the version to 16.04 LTS (Long Term Support).

Previously we were compiling ruby from source, however we have come across number of errors when older version of Ruby require additional patches to be installed. Therefore I implemented RVM (Ruby Version Manager) to manage the installation and patches.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,17 @@
-FROM ubuntu:17.04
+FROM ubuntu:16.04
 MAINTAINER development.team@moneyadviceservice.org.uk
 
-ENV BUNDLER_VERSION "1.16.0"
+ENV BUNDLER_VERSION "1.16.1"
 ENV NODE_VERSION "4.8.4"
 ENV BOWER_VERSION "1.8.2"
 ENV PHANTOMJS_VERSION "2.1.1"
-ENV PATH usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
+ENV PATH usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/rvm/bin:$PATH
 ENV RAILS_ENV test
-ENV POSTGRES postgres
 ENV BUNDLE_WITHOUT development:build
 ENV DEBIAN_FRONTEND noninteractive
 ENV APT_PACKAGES " \
-  build-essential apt-utils git libfontconfig libpq-dev libssl-dev libsqlite3-dev \
-  libxml2-dev libreadline-dev zlib1g-dev apt-transport-https curl software-properties-common"
+  build-essential apt-utils git libfontconfig libpq-dev libsqlite3-dev libmysqlclient-dev \
+  libxml2-dev libreadline-dev zlib1g-dev apt-transport-https curl software-properties-common openssh-server"
 
 #Install Prerequisites
 RUN apt-get -qq update > /dev/null && \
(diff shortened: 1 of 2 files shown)


osharpe:
The tests are currently failing in Jenkins, but this is due to errors and warnings raised by Brakeman tests.

script/test
```diff
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/bash -l
+
+source ~/.bashrc
+rvm use default
```
willhall88: Not a huge issue, but locally, because I use rbenv instead and don't have rvm then it triggers a warning about needing to download it. still seems to run the rest of the script fine

Dockerfile
```diff
@@ -25,15 +24,17 @@ WORKDIR /tmp
...
-RUN curl "https://cache.ruby-lang.org/pub/ruby/ruby-$(cat .ruby-version).tar.gz" | \
-  tar -xz && \
-  cd ruby-$(cat .ruby-version) && \
-  ./configure --enable-shared --disable-install-doc > /dev/null && \
-  make -s -j4 >> install_ruby.log && make -s install > /dev/null && \
-  rm /usr/local/lib/libruby-static.a
+
+#Download RVM as root
+RUN \curl -#LO https://rvm.io/mpapis.asc && gpg --import mpapis.asc && \
+  \curl -sSL https://get.rvm.io | bash -s stable
```
willhall88: Get the following trace when running `./script/docker_compose` - is this all normal?
```bash
Step 15/24 : RUN \curl -#LO https://rvm.io/mpapis.asc && gpg --import mpapis.asc &&   \curl -sSL https://get.rvm.io | bash -s stable
 ---> Running in 56e2f71d8a18
######################################################################## 100.0%
gpg: directory `/root/.gnupg' created
gpg: new configuration file `/root/.gnupg/gpg.conf' created
gpg: WARNING: options in `/root/.gnupg/gpg.conf' are not yet active during this run
gpg: keyring `/root/.gnupg/secring.gpg' created
gpg: keyring `/root/.gnupg/pubring.gpg' created
gpg: /root/.gnupg/trustdb.gpg: trustdb created
gpg: key D39DC0E3: public key "Michal Papis (RVM signing) <mpapis@gmail.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1  (RSA: 1)
gpg: no ultimately trusted keys found
Downloading https://github.com/rvm/rvm/archive/1.29.3.tar.gz
Downloading https://github.com/rvm/rvm/releases/download/1.29.3/1.29.3.tar.gz.asc
gpg: Signature made Sun Sep 10 20:59:21 2017 UTC using RSA key ID BF04FF17
gpg: Good signature from "Michal Papis (RVM signing) <mpapis@gmail.com>"
gpg:                 aka "Michal Papis <michal.papis@toptal.com>"
gpg:                 aka "[jpeg image of size 5015]"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 409B 6B17 96C2 7546 2A17  0311 3804 BB82 D39D C0E3
     Subkey fingerprint: 62C9 E5F4 DA30 0D94 AC36  166B E206 C29F BF04 FF17
GPG verified '/usr/local/rvm/archives/rvm-1.29.3.tgz'
Creating group 'rvm'

Installing RVM to /usr/local/rvm/
Installation of RVM in /usr/local/rvm/ is almost complete:

  * First you need to add all users that will be using rvm to 'rvm' group,
    and logout - login again, anyone using rvm will be operating with `umask u=rwx,g=rwx,o=rx`.

  * To start using RVM you need to run `source /etc/profile.d/rvm.sh`
    in all your open shell windows, in rare cases you need to reopen all shell windows.
Removing intermediate container 56e2f71d8a18
 ---> f2f3555ca52d
```

script/test
osharpe: Ye, its a good point. This test is only used for Jenkins currently, but actually it should be able to execute locally, I've added an if clause to resolve this. There is a difference when running in docker as its not interactive shell. So we need to bring in the context with the source and rvm command.

Dockerfile
osharpe: This is normal really and to be expected, we can suppress the log output which we can consider going forward. However this should only occur on the initial build of the image.

PR #423, opened 2018-01-31
[DEV APPROVED] Removed the fca import scripts as they are no longer used.

paulanthonywilson:
These are the scripts that were used when the FCA import was performed weekly by a developer.

I'm pretty sure these are no longer needed with the automated FCA import process.

What do *you* think?


margOnline:
👍

paulanthonywilson:
thank you

PR #424, opened 2018-01-31
[DEV APPROVED] 8860 Bump active_model_serializers from 0.9.x -> 0.10.x

scott-ad-riley:
`mas-rad_core` was bumped to 0.10.x ([this PR](https://github.com/moneyadviceservice/mas-rad_core/pull/164)) - which meant that it didn't need to include `self.root = false` in the serializers.

Unfortunately - bumping the gem does not mean that either `rad` or `rad_consumer` will use that version, and they are both currently on `0.9.x`.

If you do not include `self.root = false` in a serializer, in `active_model_serializers@0.9.x` - the json object returned will include the serialized model inside of a top level key, instead of just returning the serialized model.

E.g.

On 0.9.x: (this was copy pasted from a `rad` production console)
```ruby
irb(main):016:0> FirmSerializer.new(firm).as_json
=> {"firm"=>{:_id=>1306, :registered_name=>"Stafford & Co Ltd", ...
```

On 0.10.x: (from a console in this branch)
```ruby
FirmSerializer.new(firm).as_json
=> {:_id=>4, :registered_name=>"Financial Advice 5 Ltd.", ...
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ ruby '2.2.2'
 gem 'rails', '4.2.10'
 
 gem 'active_link_to'
-gem 'active_model_serializers', '0.9.4'
+gem 'active_model_serializers', '~> 0.10.1'
 gem 'azure-storage'
 gem 'bootstrap-sass'
 gem 'bowndler', github: 'moneyadviceservice/bowndler'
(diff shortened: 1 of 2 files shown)


Tr4pSt3R:
@scott-ad-riley just to get this right.. what we are interested is for the serialised model to be returned without being wrapped in a top-level key?

scott-ad-riley:
Yes

PR #425, opened 2018-01-31
[DEV APPROVED] Add firms master selection on fca import back

tomas-stefano:
[TP 8838](https://moneyadviceservice.tpondemand.com/entity/8838-rad-fca-import-report-shows-wrong)

### Bug

When you try to upload the FCA zip files to import all firms, you will see 0 (zero) firms to be imported.

### Steps to reproduce the bug:
 
Go to https://radsignup.moneyadviceservice.org.uk/admin/lookup/fca_import
Click FCA Import in Look up data
In Lookup_firms showing **after:0**

### The cause of the bug

The PR done on #415 was deleted on the next deploys, perhaps someone did a rebase by accident.
I manage to put it back but seeing the code a second time I moved to a method to make the code cleaner.

### DIFF
--- a/lib/fca/import.rb
+++ b/lib/fca/import.rb
@@ -48,7 +48,7 @@ module FCA
     def import(file)
       outcomes = River.source(context)
                       .step(&download(file))
-                      .step(&unzip(/^firms2\d+\.ext$/, /^indiv_apprvd2\d+\.ext$/, /^firm_names2\d+\.ext$/))
+                      .step(&unzip(files_to_be_imported))
                       .step(&to_sql(LOOKUP_TABLE_PREFIX))
                       .step(&save)
                       .sink
@@ -61,6 +61,10 @@ module FCA
       end
 
       [file, import_status, outcomes]
+    end
+
+    def files_to_be_imported
+      [/^firms_master_list2\d+\.ext$/, /^indiv_apprvd2\d+\.ext$/, /^firm_names2\d+\.ext$/]
     end
   end
 end
--- a/spec/lib/fca/import_spec.rb
+++ b/spec/lib/fca/import_spec.rb
@@ -51,4 +51,18 @@ RSpec.describe FCA::Import do
       end
     end
   end
+
+  describe '#files_to_be_imported' do
+    subject(:files_to_be_imported) do
+      described_class.new(files, {}, []).files_to_be_imported
+    end
+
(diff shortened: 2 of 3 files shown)


spec/lib/fca/import_spec.rb
```diff
@@ -51,4 +51,18 @@
       end
     end
   end
+
+  describe '#files_to_be_imported' do
+    subject(:files_to_be_imported) do
+      described_class.new(files, {}, []).files_to_be_imported
+    end
+
+    it 'returns regexp of the three files' do
```
Tr4pSt3R: Why no rewrite this like as: 

```ruby
     it 'returns regexp of the three files' do
        expect(files_to_be_imported).to eq([
          /^firms_master_list2\d+\.ext$/,
          /^indiv_apprvd2\d+\.ext$/,
          /^firm_names2\d+\.ext$/
        ])
    end
```

It's a matter of style but I believe the above is easier on the eye for readability.

paulanthonywilson: To be honest I'd be ok with either style.

I'd maybe favour @Tr4pSt3R myself, but not really for readability; in general the first approach is less likely to run into line-length issues.

lib/fca/import.rb
```diff
@@ -62,5 +62,9 @@ def import(file)
 
       [file, import_status, outcomes]
     end
+
+    def files_to_be_imported
```
paulanthonywilson: Moving this into its own method is great.

Just a note that it does make it harder to spot the difference when reviewing, though. :-/

spec/lib/fca/import_spec.rb
tomas-stefano: @Tr4pSt3R  I choose yours, but Travis failed [here](https://travis-ci.org/moneyadviceservice/rad/builds/335732528). Then I ran rubocop with the flag -a and rubocop autocorrected to that.

paulanthonywilson: 🤦‍♀️

margOnline:
👍

spec/lib/fca/import_spec.rb
Tr4pSt3R: Hmm, Oh that's very interesting! Well if Rubocop and Travis are in agreement then I am in agreement too. 👍

PR #426, opened 2018-03-16
[DEV APPROVED] Update Jenkins config to improve operation

osharpe:
Some simple updates to config to improve Jenkins operation

**Jenkinsfile**

- Setting JOB_NAME based on repo name
- Mapped Github OAuth credentials through the Jenkins Credentials Plugin to give the ability for Bundler to download and install private MAS Github Repo's
- Adding Post job stage to clean up workspace

**Testing**

- Update Brakeman version to 4.2.0
- Setting the Brakeman test to run as information only. Brakeman tests won't fail the Jenkins job
- Setting environment var to UTF-8 to prevent Brakeman errors
- Added --quiet arguments to npm and bundle install to reduce noise in logs

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,7 +87,7 @@ GEM
     bootstrap-sass (3.3.5)
       autoprefixer-rails (>= 5.0.0.1)
       sass (>= 3.2.19)
-    brakeman (4.1.1)
+    brakeman (4.2.0)
     builder (3.2.3)
     byebug (9.1.0)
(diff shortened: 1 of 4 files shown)


tomas-stefano:
Is brakeman failing on here? I am asking because I am not sure if we should just run as info the brakeman. If brakeman fails the Jenkins should fail.

osharpe:
Cool, I will create additional ticket to address the warning identified by Brakeman.

PR #429, opened 2018-12-18
[DEV APPROVED] - [9963] Merge `mas-rad_core` into `rad`

your:
[TP 9963](https://moneyadviceservice.tpondemand.com/entity/9963-move-mas-rad_core-into-rad-admin)

This PR is part of the work to merge `mas-rad_core` into one or two single applications.
Rationale behind can be found in the commit descriptions and/or the TP card linked above.

This PR is huge and cannot be split, here's why:

1. Being huge means a lot of files have been moved, files that prove to belong to this app indeed
2. Moving a lot of files means you need to deal with rubocop as well. `rubocop_todo.yml` has been created as well.

**Deployed to staging as of 20/12/2018: https://mas-rad-staging.herokuapp.com/**


.rubocop.yml
```diff
@@ -1,17 +1,25 @@
+inherit_from: .rubocop_todo.yml
+
```
malcolmbaig: Should we be inheriting from the mas-standards rubocop here?

tomas-stefano:
I will consider as dev approved since it has two thumbs.

.rubocop.yml
your: A fair point @malcolmbaig @tomas-stefano. Added it and updated the `.rubocop_todo.yml` file as we might need to address rubocop better in another instance.

PR #435, opened 2019-02-11
[DEV APPROVED] 9238 Make adviser details mandatory for remote firms

margOnline:
[TP 9328](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&searchPopup=userstory/9238)

### PROBLEM
Currently, when a firm only provides phone or online advice, it is not required to enter any advisers. 

image

### SOLUTION
There is now a requirement for these firms to have at least 1 adviser. The first instalment of this requirement is to show a warning and highlight where there are no advisers.

firm prompted to enter details

### DIFF
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -193,7 +193,7 @@ class Firm < ActiveRecord::Base
   end
 
   def missing_advisers?
-    (primary_advice_method == :local) && advisers.empty?
+    advisers.empty?
   end
 
   private
--- a/app/models/firm.rb
+++ b/app/models/firm.rb
@@ -189,11 +189,7 @@ class Firm < ActiveRecord::Base
   end
 
   def publishable?
-    registered? && offices.any? && !missing_advisers?
-  end
-
-  def missing_advisers?
-    advisers.empty?
+    registered? && offices.any? && advisers.any?
   end
 
   private
--- a/spec/presenters/self_service/status_presenter_spec.rb
+++ b/spec/presenters/self_service/status_presenter_spec.rb
@@ -149,7 +149,7 @@ RSpec.describe SelfService::StatusPresenter do
         let(:primary_advice_method) { :remote }
 
         it 'provides "tick"' do
-          expect(presenter.advisers_icon).to eq('tick')
+          expect(presenter.advisers_icon).to eq('exclamation')
         end
       end
 
@@ -283,7 +283,7 @@ RSpec.describe SelfService::StatusPresenter do
         let(:primary_advice_method) { :remote }
 
         it 'is false' do
-          expect(presenter.needs_advisers?).to eq(false)
+          expect(presenter.needs_advisers?).to eq(true)
         end
       end
 
--- a/spec/presenters/self_service/status_presenter_spec.rb
+++ b/spec/presenters/self_service/status_presenter_spec.rb
@@ -148,7 +148,7 @@ RSpec.describe SelfService::StatusPresenter do
       context 'and is remote' do
         let(:primary_advice_method) { :remote }
 
-        it 'provides "tick"' do
+        it 'provides "exclamation"' do
           expect(presenter.advisers_icon).to eq('exclamation')
         end
       end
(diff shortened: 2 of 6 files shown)


app/models/firm.rb
```diff
@@ -189,11 +189,7 @@ def main_office
   end
 
   def publishable?
-    registered? && offices.any? && !missing_advisers?
-  end
-
-  def missing_advisers?
-    (primary_advice_method == :local) && advisers.empty?
+    registered? && offices.any? && advisers.any?
```
your: 👏 

Just to make it clear: by changing the behaviour of the `publishable?` method, we also [control who's indexed or not (or removed from the existing index)](https://github.com/moneyadviceservice/rad/blob/0e797a10757914475efe858708cc2c207a001512/lib/firm_indexer.rb#L4-L8). That is specifically when:

1. The index is [manually (re)created](https://github.com/moneyadviceservice/rad/blob/6b9c0eca644b0d3fd9878b778cb4caf0f4fa7eef/lib/tasks/index.rake#L11) (we don't really do it apparently)
2. An update is triggered on a [firm](https://github.com/moneyadviceservice/rad/blob/0e797a10757914475efe858708cc2c207a001512/app/models/firm.rb#L113), [adviser](https://github.com/moneyadviceservice/rad/blob/0e797a10757914475efe858708cc2c207a001512/app/models/adviser.rb#L37) or [office](https://github.com/moneyadviceservice/rad/blob/0e797a10757914475efe858708cc2c207a001512/app/models/office.rb#L43) model (i.e. addition/update from the weekly FCA import _or_ manual update from the self signup area _or_ by a RAD admin) 

This should not be a problem, since it would still allow some firms/advisers to have a grace period.

I just want to highlight that: **some firms/advisers might start being removed from the index because of any of 2. happens at any point in time.** Better confirm that with product, still possibly fine and in line with what we want, but just to be extra clear about it.

margOnline: Thanks @your, I will confirm with product. The issue is that product are asking for conflicting functionality; the warning message is dependent on the `publishable?` method. 

I considered would be to use the `created_at` field to preclude firms created before a designated date from being removed from the index. This approach which would allow the onboarding warning to be displayed. But I would argue against that approach.

your: @MarcoBgn I think it should be fine and we should keep it simple, basically they will all get removed _eventually_. It's just that we cannot predict exactly _when_ at the moment if we introduce these changes (which by the way, I do approve), because several events can cause the removal. As long as product is aware of it, maybe it's just fine. Hoping they have some comms ready for those firms/advisers in case.

spec/presenters/self_service/status_presenter_spec.rb
```diff
@@ -283,7 +283,7 @@ def office_link_regex(model, text)
         let(:primary_advice_method) { :remote }
 
         it 'is false' do
-          expect(presenter.needs_advisers?).to eq(false)
+          expect(presenter.needs_advisers?).to eq(true)
```
scruti: :raising_hand_man: test description needs to be changed

spec/presenters/self_service/status_presenter_spec.rb
```diff
@@ -149,7 +149,7 @@ def firm_link_regex(model, text)
         let(:primary_advice_method) { :remote }
 
         it 'provides "tick"' do
-          expect(presenter.advisers_icon).to eq('tick')
+          expect(presenter.advisers_icon).to eq('exclamation')
```
scruti: :raising_hand_man: test description needs to be changed.

margOnline:
@your, @scruti all comments addressed, thx ♻️

app/models/firm.rb
margOnline: I have spoken to product and to the stakeholder who has confirmed this approach and will initiate the necessary comms to ensure the existing remote firms with 0 advisers update their listings. I believe this pr can progress to a test environment, as long as we don't progress to production until we get the go-ahead from the stakeholder.

PR #438, opened 2019-03-11
Bugfix: consistently upcase reference number when building adviser

malcolmbaig:
[TP](url)

Ensures adviser is built with upcased reference number, for consistent
comparisons with the lookup table advisers.

One scenario this bug occurs in is if a user tries to add an adviser to
a firm with Javascript disabled in the browser. The JS that upcases the
user input isn't available and so the server returns an error because it
can't find a matching adviser in the lookup data.


app/models/adviser.rb
```diff
@@ -10,9 +10,9 @@ class Adviser < ActiveRecord::Base
   has_and_belongs_to_many :professional_standings
   has_and_belongs_to_many :professional_bodies
 
-  before_validation :assign_name, if: :reference_number?
-
   before_validation :upcase_postcode
+  before_validation :upcase_reference_number
```
MarcoBgn: 👍

your: This will basically affect the way every new record is also inserted, which is actually good. Like it 👍

app/models/adviser.rb
```diff
@@ -95,6 +95,10 @@ def upcase_postcode
     postcode.upcase! if postcode.present?
   end
 
+  def upcase_reference_number
+    reference_number&.upcase!
```
margOnline: Like the use of the safe navigator operator which we could not use prior to the ruby upgrade!

PR #442, opened 2019-04-25
10269 - Replace Elasticsearch with Algolia

your:
[TP 10269](https://moneyadviceservice.tpondemand.com/entity/10269-rad-replace-elasticsearch-with-algolia-and)

This PR introduces firms/advisers & offices indexing in Algolia.
It's needed by https://github.com/moneyadviceservice/rad_consumer/pull/352.

Hint: best way to start is by looking at the serializers.


lib/algolia_index/base.rb
```diff
@@ -0,0 +1,61 @@
+module AlgoliaIndex
```
malcolmbaig: Minor one, but perhaps this should be in `alogolia_index/models`, given it's the model base class?

lib/algolia_index/models/firm.rb
```diff
@@ -0,0 +1,15 @@
...
+    def update!
+      advisers = object&.advisers&.geocoded
+      return if advisers.blank?
+
+      serialized = advisers.map(&AlgoliaIndex::AdviserSerializer.method(:new))
+      AlgoliaIndex.index_advisers.add_objects(serialized)
+    end
+
+    def destroy!
+      # Nothing to be directly destroyed in any index
```
malcolmbaig: Worth raising an exception of some sort then, rather than just returning nil?

lib/algolia_index.rb
```diff
@@ -0,0 +1,37 @@
...
+        firm_id: firm_id
+      )
+      if record.exists?
+        record.update!
+      else
+        record.destroy!
+      end
+    end
+
+    def index_advisers
```
malcolmbaig: Perhaps `indexed_advisers` instead - reads more like a noun that way, current form reads like a verb.

Same for offices below.

lib/algolia_index/models/firm.rb
tomas-stefano: This is a very very good comment. Although there are some objects that call then and in some should be doing something and others not. Based on this comment, me and @your are thinking to rename to just `destroy` (without a bang) so that the return nil make sense. Wdyt?

lib/algolia_index.rb
your: @malcolmbaig that's true! I think at some point we were using  `advisers_index` and `offices_index` but somehow we ended reversing the order.

lib/algolia_index/models/firm.rb
your: @malcolmbaig That's a very good comment, thanks for raising it.

Since this method is triggered anyway when a firm is destroy, it should not do anything, as we don't have a firm index.

But whenever a firm is destroyed in the db, their advisers and offices will be destroyed because of the cascade, which will trigger the removal of advisers and offices from their relative indices. 

So this method here, should not do anything really, but it's very confusing to have the banged version of it when it does not raise indeed, so I think we could just remove the bang and avoid confusion.

What do you think, does it make sense?

lib/algolia_index.rb
```diff
@@ -0,0 +1,37 @@
...
+  }.freeze
+
+  class << self
+    def handle_update!(klass:, id:, firm_id:)
+      record = "AlgoliaIndex::#{klass}".constantize.new(
+        klass: klass,
+        id: id,
+        firm_id: firm_id
+      )
+      if record.exists?
```
malcolmbaig: This internal API is slightly confusing unless you read the implementation. I think it would help to make some of these variables/method names more explicit. For example:

`search_record = "AlgoliaIndex::#{klass}".constantize.new`

`if search_record.db_record_exists?`

lib/algolia_index/base.rb
```diff
@@ -0,0 +1,61 @@
...
+      @object ||= klass.constantize.find_by(id: id)
+    end
+
+    def validate_init
+      instance_klass = self.class.name.split('::').last
+      raise ObjectClassError if instance_klass == 'Base'
+      return if instance_klass == klass
+
+      raise ObjectClassError.new(expected: instance_klass, got: klass)
+    end
```
malcolmbaig: Is this constant checking code necessary? What kind of scenario would prompt an error of this sort?

lib/algolia_index.rb
your: @malcolmbaig Good point, glad you got the difference in type of record and proposed an alternative naming.

lib/algolia_index/base.rb
your: @malcolmbaig Think it's just me being pedantic. Not really needed, but you could by accident pass any AR object and for instance remove and office with the id of an adviser instead. Very unlikely, might remove it as well.

lib/algolia_index/models/firm.rb
malcolmbaig: I think that makes sense - I'd take it a step further and remove the bangs from other methods as well. After removing from destroy, we no longer have a consistent meaning for `!` across these objects so I don't think it adds clarity, IMHO. 

A slightly more detailed comment would help too I reckon, perhaps something like: 

`# Do nothing, because we don't maintain a Firm index.`

your: @malcolmbaig Totally, needs to be removed from all of them. Will rephrase the comment if that's clearer. 👍

lib/algolia_index.rb
tomas-stefano: ♻️ :)

lib/algolia_index/base.rb
tomas-stefano: ♻️ :)

lib/algolia_index.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module AlgoliaIndex
+  INDICES = {
+    advisers: "firm-advisers#{'-test' if Rails.env.test?}",
+    offices: "firm-offices#{'-test' if Rails.env.test?}"
+  }.freeze
+
+  class << self
+    def handle_update(klass:, id:, firm_id:)
+      index = "AlgoliaIndex::#{klass}".constantize.new(
```
malcolmbaig: Still think this is something more like 'indexed_record' than 'index'. WDYT?

malcolmbaig: PR's looking good btw, only have this question left.

tomas-stefano: Looks good to me 👍

tomas-stefano: ♻️ done! :)

PR #443, opened 2019-04-26
[DEV APPROVED] 10498 - Principal unable to sign in (bug)

your:
[TP 10498](https://moneyadviceservice.tpondemand.com/entity/10498-rad-principal-unable-to-sign-in)

This PR's goal is to prevent the re-occurrence of the bug described above (in short: a principal existed but the associated user did not).

Please find more information attached to the card including the comments and the commit description.

### DIFF
--- a/app/controllers/admin/principals_controller.rb
+++ b/app/controllers/admin/principals_controller.rb
@@ -28,8 +28,10 @@ class Admin::PrincipalsController < Admin::ApplicationController
 
     message = destroy_message(principal)
 
-    user.principal.destroy
-    user.destroy
+    ActiveRecord::Base.transaction do
+      user.principal.destroy!
+      user.destroy!
+    end
 
     redirect_to admin_principals_path, notice: message
   end
--- a/spec/controllers/admin/principals_controller_spec.rb
+++ b/spec/controllers/admin/principals_controller_spec.rb
@@ -1,0 +1,42 @@
+RSpec.describe Admin::PrincipalsController, type: :controller do
+  before { sign_in(user) }
+
+  let(:user) { FactoryGirl.create(:user, principal: principal) }
+  let(:principal) { FactoryGirl.create(:principal) }
+
+  describe 'DELETE #destroy' do
+    context 'when successful' do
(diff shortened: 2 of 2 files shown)


spec/controllers/admin/principals_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+RSpec.describe Admin::PrincipalsController, type: :controller do
+  before { sign_in(user) }
+
+  let(:user) { FactoryGirl.create(:user, principal: principal) }
```
margOnline: Is this an opportunity to upgrade to FactoryBot?

app/controllers/admin/principals_controller.rb
```diff
@@ -28,8 +28,10 @@ def destroy
 
     message = destroy_message(principal)
 
-    user.principal.destroy
-    user.destroy
+    ActiveRecord::Base.transaction do
+      user.principal.destroy!
+      user.destroy!
+    end
```
margOnline: Nice!

spec/controllers/admin/principals_controller_spec.rb
your: @margOnline good point, unfortunately I have a PR that is a bit chunky on the way, so it would cause me some rebase issues I'd rather avoid now, definitely worth adding a card to the tech debt backlog though.

app/controllers/admin/principals_controller.rb
MarcoBgn: Is there any reason why we cannot let Rails take care of the ref integrity as it's done here:
https://github.com/moneyadviceservice/rad/blob/master/app/models/principal.rb#L11

And just call `principal.destroy`?

your:
@MarcoBgn that's a good point, I have mentioned my rationale _against_ letting the db handle it (which I am an advocate of), possibly just a fear. But I don't know if the the relationship cardinality between `Principal` and `User` is a `1..1` or a `1..(0..1)`. In other words, I don't want to have a `User` removed unless explicitly done, haven't investigated much around the rest of the codebase. I would otherwise do exactly what you suggested.

PR #445, opened 2019-06-03
10324 Update adviser form

malcolmbaig:
[TP](https://moneyadviceservice.tpondemand.com/entity/10324-rad-edit-adviser-details-screen)

A series of data changes related to the Qualifications, Statements of Professional Standing, and Professional Body Memberships that are selectable on the adviser edit screen.

Each commit contains one discrete change in the series, so reviewing by commit highly recommended.

Post-deploy checklist:

- Run the rake task in https://github.com/moneyadviceservice/rad/pull/445/commits/e9a7bace88562419f5603b27c365b4cf39890b13.
- Re-index.
- Merge a separate PR that drops professional_bodies (see https://github.com/moneyadviceservice/rad/pull/445/commits/f08b7e5f3f814fde3e7cad067aedd6f5c0ed686f)


db/migrate/20190523124109_update_qualification_names.rb
```diff
@@ -0,0 +1,25 @@
...
+      },
+      {
+        old: 'Certified Financial Planner',
+        new: 'Certified Financial Planner - accredited by the Chartered Institute of Securities and Investments'
+      }
+    ]
+
+  def up
+    NAMES.each do |name|
+      Qualification.where(name: name[:old]).first!.update(name: name[:new])
```
your: Isn't it worth banging the update? How else would we realise it actually failed otherwise?

lib/data_migrations/merge_ifp_with_cis.rb
```diff
@@ -0,0 +1,65 @@
+class MergeIfpWithCis
+  def run_migration
+    remove_old_body_selections(adviser_ids_where_both_bodies_selected)
+    update_old_body_selections(adviser_ids_where_only_one_body_selected)
```
your: Is it worth printing some details from these mutations? Might be useful for debugging purposes as we cannot have a "migrate down" sort of thing.

malcolmbaig: Yeah, I think you're right. On a related note, I've just realised these operations are going to run on 3 and  ~160 records, respectively. Completely over estimated the target dataset size. Given the low numbers, I'm going to replace some of this SQL with simple iteration and make the updates via the Adviser model. Gives an opportunity to add some logging about the data change.

PR #447, opened 2019-06-05
[WIP] 10527 Use FCA API to verify new firms

margOnline:
[TP 10527](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=userstory/10537&searchPopup=userstory/10527)

BACKGROUND
When a new firm is created, we verify its existence against a 'Look Up' table. That table is created from files imported from the FCA. At some point, the FCA will discontinue support for these files and replace them with its API.

REQUIREMENT
Ensure a new firm is FCA authorised before it is published on RAD. Verify the reference number by checking for its existence on the FCA API, instead of checking the Look up table.

TECHNICAL / NOTES FOR REVIEWING
The majority of the files changed are to fix broken tests. These tests use a Firm Factory as part of what is being tested. Since a firm's validation now depends on an api request to an external source, that call needs to be stubbed. There are 2 approaches I considered for stubbing the api request:
1. The one here, where I created a shared context and included it in all the test files that needed it.
2. Creating a before hook in the spec_helper to stub the api request and response, you can view it on this branch, https://github.com/moneyadviceservice/rad/tree/global_stub_api_call

I chose the first approach because it is more explicit that a dependency exists between many of the tests and a valid Principal which includes an api request. 

The second approach hides this dependency. All tests use it even if they don't require it, which doesn't seem right to me. Also The FcaApi Request spec was failing because the api call was happening too early. Hence the expectation in the test for the call to occur failed, because it had already happened. There could be a work around to the failure, but that seems wrong to hack a test which is supposed to be testing the new feature.

However, I am open to other opinions. The point of opening for review this pr at this time is to get others' opinions. I am aware that there is a failing test, which of course will be fixed, but now is a good time to get a review on the approach.

Use the commit history to review this pr so you can separate the test fixes from the new code for implementing the api call.

BEFORE DEPLOYMENT
- [ ] Add the environment variables


spec/controllers/admin/firms_controller_spec.rb
```diff
@@ -10,6 +10,12 @@ def create_user_with_firm(user_attrs, firm_attrs = {})
   end
 
   describe '#login_report' do
+    before do
+      allow_any_instance_of(FcaApi::Request)
+        .to receive(:get_firm)
+        .and_return(instance_double(FcaApi::Response, ok?: true))
+    end
+
```
scruti: Noticing the amount of times and places we need to declare this `FcaApi` stub I wonder if we should stub it globally with a shared context. Something like this in the global RSpec config so it becomes a default:

```
RSpec.configure do |rspec|
  rspec.include_context "fca_api_ok_response", :include_shared => true
end
```
With this we would just need to override the default stub when we want to test the specific scenario when then API responds that the firm wasn't found.

Also think we can transform the stub to avoid the `allow_any_instance_of` that it tends to be discouraged.
I think something like this would achieve the same results (`let` declarations could be extracted if the stub inside stub approach became confusing):

```
before do
  allow(FcaApi::Request).to receive(:new).and_return(
    instance_double(
      FcaApi::Request, get_firm: instance_double(FcaApi::Response, ok?: true)
    )
  )
end
```

Didn't test any of this so I may be wrong, but hope the idea its clear. Let me know your thoughts :)

app/models/principal.rb
```diff
@@ -92,7 +92,7 @@ def associate_firm
   end
 
   def match_fca_number
-    unless Lookup::Firm.exists?(fca_number: fca_number)
+    unless fca_authorised_firm?(fca_number)
```
tomas-stefano: What would happen if the FCA call fails? Error page for the user or an error message?

db/schema.rb
```diff
@@ -176,6 +177,27 @@
...
+    t.datetime "updated_at",                               null: false
+  end
+
+  create_table "last_week_lookup_subsidiaries", force: :cascade do |t|
+    t.integer  "fca_number",                          null: false
+    t.string   "name",       limit: 255, default: "", null: false
+    t.datetime "created_at",                          null: false
+    t.datetime "updated_at",                          null: false
+  end
+
```
tomas-stefano: Are this lines supposed to be committed?

margOnline:
@tomas-stefano thanks for the review. This pr is the first user story of the [feature to replace the import files with the api](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=feature/10522&appConfig=eyJhY2lkIjoiQzE1REE4NkE5M0U2Q0VFM0JCM0QyMzM2RUM0MzBGOUQifQ==). The final story which is dependent on all other stories being completed is to remove the code associated with file import and the lookup tables.

margOnline:
Closing this pr in favour of #454

PR #450, opened 2019-06-13
[DEV APPROVED] 10544 Add two qualifications

malcolmbaig:
[TP](https://moneyadviceservice.tpondemand.com/entity/10544-rad-add-two-adviser-qualifications)

- Write a migration that adds two new qualifications and updates the
  order of the list.
- Update the database seeds with these qualifications.


malcolmbaig:
Thanks @tomas-stefano @scruti

PR #451, opened 2019-06-14
[DEV APPROVED] 10522 FCA Api setup

margOnline:
[TP 10522](https://moneyadviceservice.tpondemand.com/restui/board.aspx#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&boardPopup=feature/10522)

The requirement is to use the FCA API to verify firms and advisers. This pr sets up the connection to use the api for the various purposes as defined in the TP feature.


lib/fca_api/connection.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(domain)
+      @domain = domain
+      @api_email = ENV.fetch('FCA_API_EMAIL')
+      @api_key = ENV.fetch('FCA_API_KEY')
+      @retries = ENV.fetch('FCA_API_MAX_RETRIES').to_i
+      @timeout = ENV.fetch('FCA_API_TIMEOUT').to_i
+      configure_raw_connection
+    end
+
+    def configure_raw_connection
```
scruti: do we need to expose this method in the Connection public interface?

lib/fca_api/response.rb
```diff
@@ -0,0 +1,15 @@
...
+    SUCCESS_MESSAGE = 'ok'
+      
+    attr_reader :response
+
+    def initialize(response)
+      @response = response
+    end
+
+    def ok?
+      response.body['Message'].downcase.include?(SUCCESS_MESSAGE)
```
scruti: just to double check.... Do API responses always include a body with a 'Message'? Trying to avoid situations where the API does not return what we expect and instead of a controlled `ok? false` response it blows up :sweat_smile:

margOnline: For the firm endpoint yes. I was trying to not pre-empt future uses as it's more likely that with additional endpoints we'll have a better idea of how to structure the Response class.

lib/fca_api/connection.rb
margOnline: good point, have privatised it.

PR #454, opened 2019-06-20
[WIP] - 10527 Verify FCA number using the FCA API

margOnline:
[TP 10527](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&searchPopup=userstory/10527)

BACKGROUND
When a new firm is created, we verify its existence against a 'Look Up' table. That table is created from files imported from the FCA. At some point, the FCA will discontinue support for these files and replace them with its API.

REQUIREMENTS
Ensure a new firm is FCA authorised before it is published on RAD. Verify the reference number by checking for its existence on the FCA API, instead of checking the Look up table.

If the fca number does not exist on the FCA Register or the FCA API is not available, we still create the principal and set a flag to indicate the fca number needs verification. The principal is highlighted on the admin section which enables admin users to see on the index page quickly that manual intervention is required. 

On the show page, the admin user can view the full details to check the fca number against their own list, and click a button to verify the fca number. 

TECHNICAL
I had to change how the firm_search_spec features are tested as they were using Firms and with this feature, we introduce a change which requires Firms to have an associated Principal to adequately test these specific features. The creation of Firms to test various search features can probably be improved and I looked at a number of approaches using various FactoryGirl features such as traits, parent, association and after methods but they didn't work or perhaps I applied them incorrectly. As always, open to any and all comments etc.

### DIFF
--- a/spec/features/admin/firm_approval_spec.rb
+++ b/spec/features/admin/firm_approval_spec.rb
@@ -1,18 +1,22 @@
 RSpec.feature 'Approving firms on the admin interface', :inline_job_queue do
   include ActiveSupport::Testing::TimeHelpers
+  include_context 'fca api ok response'
   include_context 'algolia directory double'
 
   let(:index_page) { Admin::FirmsIndexPage.new }
   let(:firm_page) { Admin::FirmPage.new }
   let(:approval_date) { Time.utc(2019, 6, 1) }
 
-  scenario 'Approving a Firm' do
+  
+  scenario 'Approving an fca verified Firm' do
     given_there_are_fully_registered_principal_users_with_advisers_and_offices
     given_i_am_at_the_admin_firms_index_page
     then_i_see_all_firms
     then_all_firms_are_not_approved
     and_no_advisers_or_offices_are_present_in_the_directory
 
+    when_i_visit_a_firm_page(@firms.first)
+    when_i_verify_the_firm
     when_i_visit_a_firm_page(@firms.first)
     then_i_see_the_firm_is_not_approved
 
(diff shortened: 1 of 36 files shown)


spec/features/admin/firm_approval_spec.rb
```diff
@@ -1,18 +1,22 @@
 RSpec.feature 'Approving firms on the admin interface', :inline_job_queue do
```
malcolmbaig: This spec is failing when I run this branch locally, and I think it's a pretty good illustration of some of the issues we're introducing with these changes.

Primarily, this spec doesn't really accurately reflect the behaviour of the system anymore. If we have "fully_registered_principal_users", the way the code's written means that we expect an API-based verification attempt to be triggered in a callback at the point that a principal is created. In switching off the TestAfterCommit config (elsewhere in this PR), two things have happened:

- The code under test doesn't behave like it does in non-test environments unless we explicitly execute the after_commit callbacks.
- We're relying on the fact that those callbacks aren't run to make sure that the database is in the right state for us to be able to click a verify button manually later on in the scenario. This is very hidden and makes following the test harder.

These issues are compounded by the fact that the `verify_fca_number` callback added to `Principal` creates an infinite loop in its current form, and needs to changed to update the column directly (or some other solution).

I'm going to have a crack at reworking some of this to address the issues above, will commit directly to this branch.

margOnline: @malcolmbaig good shout

margOnline:
Closing this pr in favour of a new approach which will be to query the api as a background job.

PR #455, opened 2019-06-21
Add Algolia directory feature tests

scruti:
[TP-10560](https://moneyadviceservice.tpondemand.com/entity/10560-add-expectations-over-directory-contentupdates-in)

Current integration tests cover the management of
Principals, Firms, Advisers and Offices from both the Principals and
Admins user flows.

The test coverage is missing a key part of RAD business logic though:
"What and when gets pushed/removed to the RAD Directory?"

The main purpose of RAD, besides allow Principals/Admins to maintain
firms advisers and offices, is to provide that data to the public
consuming Retirement Adviser Directory. And that should be covered
with integration tests.

The retirement adviser directory publicly accessible information gets
pushed from this app to Algolia. Containing:
- One index for firm Advisers.
- One index for firm Offices.

Being able to inspect what the index contains at the end of user flows
is quite tricky, and can have different approaches:

(TLDR: Jump to the fifth Approach - We used an Algolia Test Double)

### First approach: Use real Algolia cloud stored test indexes.

We have shared Algolia test indexes running exactly as in production.
The problem is that this data is shared between different test/runs and
data from one test can pollute the expectations of other tests.

On top of the test isolation issue, we would have a real dependency on
the ALGOLIA API KEY for testing and we would need to hit
Algolia service multiple times for each feature test run.

### Second approach: Set expectations over Webmock/VCR cassetes.

This would cover when we want to check that "this object gets pushed to
Algolia" or "this object gets removed from Algolia", but it would imply
a quite big number of cassettes and also doesn't make easy to just check:
"Algolia Index contains X but not Y" expectations at the end of an user
flow.

### Third approach: Use Algolia provided testing Mocks.

Algolia gem provides Webmock stubs collection for testing purposes.
Problem is that those stubs return fixed data, with fixed ID and
information. If we want to test what do we actually push/delete from
our indexes, this approach wouldn't allow us, as indexes will always
return the same unrelated dummy data.

### Fourth approach: Just set expectations over calls on Algolia Indexes.

Classic "I expect Algolia to receive a call over add element to the
index".

This stubbing works great for Unit testing, when you isolate bits of
business logic and just expect "X to be called", but we are doing
end to end feature testing, and we mainly want to check "What Algolia
contains after this user flow happen".
In order to get anything similar
to that setting call expectations over Algolia Index stubs, we would
need to know too much about the implementation and define lots of
stubs, making tests quite brittle and coupled with the implementation.

### Fifth approach: Replace real Algolia for a local test double.
We don't want to be stubbing all the Algolia calls, we don't want to
use real Algolia service with network dependency and lack of test env
isolation... what do we have left?
We can just replace Algolia all together with a fake test directory
providing the same interface as Algolia but just storing the indexes
into an array.

This allow us to:
- Not have any network/api dependency.
- Test end to end what indexes contain at the end of an user flow.
- Minimal use of Stubs (we only stub the calls for instantiating a
  new Algolia::Index object).
- Reduce the coupling with implementation details.
  As a single spec context file contains all the Algolia::Index stub
  and helper methods to query the index from tests.
  If something happens to change in real Algolia gem interface/service,
  we basically have a single point to maintain, instead having
  implementation details polluting all our test files.

After trying/examining all the options, went for the fifth/last
approach, as allows us to test what we want to test with minimum
pain.

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -43,7 +43,7 @@ class Office < ActiveRecord::Base
   after_commit :notify_indexer
 
   def notify_indexer
-    UpdateAlgoliaIndexJob.perform_async(model_name.name, id, firm_id)
+    UpdateAlgoliaIndexJob.perform_later(model_name.name, id, firm_id)
   end
 
   def field_order
--- a/Gemfile
+++ b/Gemfile
@@ -86,6 +86,7 @@ group :test do
   gem 'rspec-collection_matchers'
   gem 'rspec-sidekiq'
   gem 'site_prism'
+  gem 'test_after_commit'
   gem 'tzinfo-data'
   gem 'vcr'
   gem 'webmock'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -437,6 +437,8 @@ GEM
     statsd-ruby (1.4.0)
     terminal-table (1.8.0)
       unicode-display_width (~> 1.1, >= 1.1.1)
+    test_after_commit (1.2.2)
+      activerecord (>= 3.2, < 5.0)
     thor (0.20.3)
     thread_safe (0.3.6)
     tilt (1.4.1)
@@ -530,6 +532,7 @@ DEPENDENCIES
   site_prism
   slack-ruby-client
   statsd-ruby (~> 1.4.0)
+  test_after_commit
   timecop
   tzinfo-data
   uglifier (>= 1.3.0)
--- a/app/models/adviser.rb
+++ b/app/models/adviser.rb
@@ -35,7 +35,7 @@ class Adviser < ActiveRecord::Base
   after_commit :notify_indexer
 
   def notify_indexer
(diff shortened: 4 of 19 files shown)


app/models/office.rb
```diff
@@ -43,7 +43,7 @@ class Office < ActiveRecord::Base
   after_commit :notify_indexer
 
   def notify_indexer
-    UpdateAlgoliaIndexJob.perform_async(model_name.name, id, firm_id)
+    UpdateAlgoliaIndexJob.perform_later(model_name.name, id, firm_id)
```
tomas-stefano: Just saying Margo's PR #454 also changed this line. Just saying :)

scruti: yeah we were syncing on how to fix multiple test issues :sweat_smile: . Will be rebased against this (if this gets in first as I guess). Thanks for the check!

PR #456, opened 2019-06-26
[DEV APPROVED] [TP-10531] require admin approval before publishing new firms in the directory

scruti:
[TP-10531](https://moneyadviceservice.tpondemand.com/entity/10531-rad-require-manual-admin-approval-before)

## What this PR does?

This PR:
- Notifies RAD admins by email when a new firm is registered through a new Principal signup.
- Requires the firms to be manually approved by RAD admins before they're pushed into the directory.
- Provides the RAD admin an interface to approve firms in the firm admin page.

## How do I review it?
Commit by commit helps to check the individual logic changes.

## Things to Highlight

- Once the migration runs in production, all the existent firms will be approved during the migration. So only the new firms added after this feature deploy will require a manual approval.
- Firms cannot be "unnaproved". Admins have tools to easily delete the firm though. 

## Implementation Highlights 

- Individual elements in Algolia Adviser Index contain a subset of firm information with info like `total_advisers` and `total_offices` in each adviser entry. This is why after each update on any office or adviser we need to regenerate all the firm advisers information in the index.

- We had to introduce a way to, once the admin approves a firm, trigger the push to Algolia of both advisers and offices of the firm, as they weren't in the directory while waiting for approval.  This has been done through the `AlgoliaIndex::Firm#update` method. While its previous logic for refreshing the advisers has been moved to a more explicit `AlgoliaIndex::Firm#update_advisers` 

### Email received by admins once a Principal Signs up and the firms gets associated to it
Screenshot from 2019-06-04 09-42-12

### Admin firms index interface displaying approval status
Screenshot from 2019-06-11 16-02-13

### Firm admin page enabling the firm to be approved
Screenshot from 2019-06-11 16-03-57

### Firm admin page after approving the firm displaying approval data.
Screenshot from 2019-06-11 16-02-31

### DIFF
--- a/app/views/admin/firms/show.html.erb
+++ b/app/views/admin/firms/show.html.erb
@@ -28,6 +28,14 @@
     <p>
       <strong>Approved:</strong> <%= @firm.approved_at&.to_s(:long) || 'Not approved' %>
     </p>
+    <% if @firm.approved_at.blank? %>
+      <p>
+        <%= button_to('Approve Firm',
+                      admin_firm_approve_path(@firm.id),
+                      class: 'btn btn-success',
+                      method: :post) %>
+      </p>
+    <% end %>
   </div>
 </div>
 
--- a/app/views/admin/firms/show.html.erb
+++ b/app/views/admin/firms/show.html.erb
@@ -25,6 +25,9 @@
     <p>
       <strong>Added:</strong> <%= @firm.created_at.to_s(:long) %>
     </p>
+    <p>
+      <strong>Approved:</strong> <%= @firm.approved_at&.to_s(:long) || 'Not approved' %>
+    </p>
   </div>
 </div>
 
--- a/spec/features/admin/firm_approval_spec.rb
+++ b/spec/features/admin/firm_approval_spec.rb
@@ -1,0 +1,84 @@
+RSpec.feature 'Approving firms on the admin interface' do
+  include ActiveSupport::Testing::TimeHelpers
+
+  let(:index_page) { Admin::FirmsIndexPage.new }
+  let(:firm_page) { Admin::FirmPage.new }
+  let(:approval_date) { Time.utc(2019, 6, 1) }
+
+  scenario 'Approving a Firm' do
+    given_there_are_fully_registered_principal_users
+    given_i_am_at_the_admin_firms_index_page
+    then_i_see_all_firms
+    then_all_firms_are_not_approved
+    when_i_visit_a_firm_page(@firms.first)
+    then_i_see_the_firm_is_not_approved
+
+    travel_to(approval_date) do
+      when_i_approve_the_firm
+    end
+
+    then_the_firm_becomes_approved
+    then_i_dont_see_the_approve_button
+    then_the_firm_is_listed_as_approved_in_the_index(@firms.first)
+  end
+
+  def given_i_am_at_the_admin_firms_index_page
+    index_page.load
+    expect(index_page).to be_displayed
+    expect_no_errors
+  end
+
+  def given_there_are_fully_registered_principal_users
+    @users = FactoryGirl.create_list(:user, 2)
+    @principals = @users.map(&:principal)
+    @firms = @principals.map(&:firm)
+  end
+
+  def then_i_see_all_firms
+    expect(index_page.total_firms).to eq(@firms.count)
+    expect(index_page.firms.count).to eq(@firms.count)
+  end
+
+  def then_all_firms_are_not_approved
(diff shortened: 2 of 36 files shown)


spec/features/admin/firm_approval_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    directory_offices = firm_offices_in_directory(firm)
+
+    aggregate_failures 'firm info in directory' do
+      expect(directory_advisers.map { |adviser| adviser['objectID'] })
+        .to eq firm.advisers.pluck(:id)
+      expect(directory_offices.map { |office| office['objectID'] })
+        .to eq firm.offices.pluck(:id)
+    end
+  end
+end
```
margOnline: Nice use of `rspec_all` and `travel_to`! I'm not familiar with them so its good to learn about them.

spec/support/admin/firm_page.rb
```diff
@@ -7,4 +7,6 @@ class Admin::FirmPage < SitePrism::Page
   element :move_advisers, '.t-move-advisers'
   element :sign_in_as_principal, '.t-sign-in-as-principal'
   element :new_adviser, '.t-new-adviser'
+  element :approved, '.t-approved'
+  element :approve_button, '.t-approve-firm'
```
margOnline: I don't like adding classes to the markup to facilitate testing because it adds unnecessary markup.  I know that the standard on this project has been to use these `t-` classes but if it can be avoided easily, I think we should try to. Is it possible to do something like
```
element :approved_elem, 'input[value="Approve Firm"]'
```

spec/support/admin/firm_page.rb
```diff
@@ -7,4 +7,6 @@ class Admin::FirmPage < SitePrism::Page
   element :move_advisers, '.t-move-advisers'
   element :sign_in_as_principal, '.t-sign-in-as-principal'
   element :new_adviser, '.t-new-adviser'
+  element :approved, '.t-approved'
```
margOnline: same comment as the line below, tho this one may be a bit harder, perhaps using `:last-child`?

spec/support/admin/firms_index_page.rb
```diff
@@ -5,6 +5,7 @@ class Admin::FirmsIndexPage < SitePrism::Page
   class RowSection < SitePrism::Section
     element :fca_number_elem, '.t-fca-number'
     element :registered_name_elem, '.t-registered-name'
+    element :approved_elem, '.t-approved'
```
margOnline: same comment re the use of test classes

app/views/admin/firms/show.html.erb
```diff
@@ -25,6 +25,16 @@
...
+    <p class="t-approved">
+      <strong>Approved:</strong> <%= @firm.approved_at&.to_s(:long) || 'Not approved' %>
+    </p>
+    <p>
+      <%= button_to('Approve Firm',
+                    admin_firm_approve_path(@firm.id),
+                    class: 'btn btn-success t-approve-firm',
+                    method: :post,
+                    disabled: @firm.approved_at.present?) %>
+    </p>
```
margOnline: Would the ui work better if the button did not show at all if the firm has already been approved?

spec/support/admin/firm_page.rb
scruti: will give I a try. Thanks for the suggestion!

app/views/admin/firms/show.html.erb
scruti: Very good idea. Yes, going to change it.

scruti:
@margOnline suggestions addressed! Thanks for the comments!

spec/support/admin/firm_page.rb
```diff
@@ -7,4 +7,6 @@ class Admin::FirmPage < SitePrism::Page
   element :move_advisers, '.t-move-advisers'
   element :sign_in_as_principal, '.t-sign-in-as-principal'
   element :new_adviser, '.t-new-adviser'
+  element :approved, 'p:contains("Approved:")'
```
margOnline: nice!

app/views/admin/firms/show.html.erb
```diff
@@ -25,6 +25,17 @@
     <p>
       <strong>Added:</strong> <%= @firm.created_at.to_s(:long) %>
     </p>
+    <p class="t-approved">
```
margOnline: You can now remove this class

scruti: wops!

PR #461, opened 2019-08-08
[DEV APPROVED] 10592 Use the FCA API to verify new principals

margOnline:
[TP 10592](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&searchPopup=userstory/10592)

See individual commit messages for a description of the changes and approach.


app/controllers/principals_controller.rb
```diff
@@ -72,4 +55,18 @@ def new_principal_form_params
...
+    @form.as_json.slice(
+      "fca_number",
+      "first_name",
+      "last_name",
+      "job_title",
+      "email",
+      "telephone_number",
+      "password",
+      "password_confirmation",
+      "confirmed_disclaimer"
```
tomas-stefano: single quotes?

app/mailers/failed_registration_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class FailedRegistrationMailer < ActionMailer::Base
+  default from: 'RADenquiries@moneyadviceservice.org.uk'
```
tomas-stefano: Shall we move this values to an env var? What are your thoughts about it?

app/jobs/verify_reference_number_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  sidekiq_options retry: 5, backtrace: true, unique: :until_executed
+
+  def perform(form_data)
+    @response = FcaApi::Request.new.get_firm(form_data['fca_number'])
+
+    if @response.ok?
+      Rails.logger.info("\nINFO: FCA API successful request for #{form_data['fca_number']}\n")
+      VerifiedPrincipal.new(form_data, firm_name).register!
+    else
+      Rails.logger.warn("\n WARN: FCA API failed for #{form_data['fca_number']}\n")
```
tomas-stefano: Love the logs ❤️

app/models/verified_principal.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_accessor :form, :firm_name
+
+  def initialize(form_data, firm_name)
+    @form = NewPrincipalForm.new(form_data)
+    @firm_name = firm_name
+  end
+
+  def register!
+    create_new_principal
+    @user.principal.firm = create_associate_firm
```
tomas-stefano: If the principal has the primary key of fca number and firm the foreign key of fca number, how relevant is the assign of `@user.principal.firm= `? Can we not just call the `create_associate_firm`?

app/views/failed_registration_mailer/notify.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Retirement Adviser Directory Registration</h1>
+<div>
+  We haven't been able to verify your Firm Reference Number against the FCA register. Please contact us at radenquiries@moneyadviceservice.org.uk and we will help you with your registration.
```
tomas-stefano: If we setup the env var we can use and replace "radenquiries@moneyadviceservice.org.uk". An idea...

spec/models/verified_principal_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expect(user).to receive(:save!)
+      expect(Stats).to receive(:increment).with('radsignup.principal.created')
+      expect(Identification)
+        .to receive(:contact)
+        .with(user.principal)
+        .and_return(identification)
+      expect(identification).to receive(:deliver_later)
+      expect(NewFirmMailer).to receive_message_chain(:notify, :deliver_later)
+
+      subject.register!
```
tomas-stefano: There are something about this test that sounds like a smell to me, I think there are too much setup and stubs. I don't have any suggestion but shall we pair on this to try to refactor the method?

spec/support/vcr.rb
```diff
@@ -24,4 +24,8 @@
       request.proceed
     end
   end
+
+  config.ignore_request do |req|
+    req.uri.include?('register.fca') || req.uri.include?('algolia.net')
+  end
```
tomas-stefano: This lines were added before we fixed the problem of mocking the API, is these lines necessary yet?

spec/models/verified_principal_spec.rb
tomas-stefano: As I talked in person, I don't have any suggestion about this. I will left to the author take the decision on the path to take about this. 🙏

app/mailers/failed_registration_mailer.rb
margOnline: The email address is actually used in many places and to implement it as an env var is a big change which I think would detract from this feature. I will create a TP card to consider your suggestion.

app/views/failed_registration_mailer/notify.html.erb
margOnline: Please refer to the previous comment about the env var.

spec/models/verified_principal_spec.rb
margOnline: I'm of 2 minds about the comment. I understand why you have made the comment about the `code smell` due to the test setup, but I also am not sure if refactoring will provide any benefit at this stage. It may be that as we use the api more, we find a better case for refactoring. Since there has been no suggestion about how to refactor or follow up on pairing, I'm going to leave it as it is.

PR #464, opened 2019-09-26
[DEV APPROVED] - 10525 Batch Status Check Firms

malcolmbaig:
[TP](https://maps.tpondemand.com/entity/10525-rad-firm-status-check-via-api)

This PR adds functionality that checks the status of all firms against the FCA Register API and displays all firms with an inactive status in a table. The intention is to run this process in production for several days and review the results it produces. The results table is hidden from the UI for now while this feature is still in beta. I've also elected to skip tests in this implementation as:

- the feature had a relatively small footprint.
- it will likely be reworked after some time running in production.
- it's internal-facing and will be running daily - any issues/bugs will be ironed out from multiple production runs.

The API has a rate limit of 10 requests per 10 seconds. The approach here is to use Sidekiq jobs to concurrently execute these requests, and to enqueue these jobs on a schedule that throttles the number of requests executed in any 10 second window. Please refer to the comment in EnqueueFirmStatusChecksJob for more detail.

There are other approaches this solution could have taken, including:

- Specifying a throttle within the Sidekiq server, via Sidekiq Enterprise or an add-on gem.
  - Pros: no scheduling code, jobs can be enqueued and are throttled automatically.
  - Cons: nothing obvious.
- A single, long-running Ruby process using threads.
  - Pros: easily report start and finish of the entire batch process.
  - Cons: would need to write retry mechanism and error handling from scratch, process crash would terminate all work.

Of these, the better alternative is the first one. If the solution in this PR doesn't hold up well in production, we can rework it along those lines. This is dependent on updating our Sidekiq version in RAD, however.

### DIFF
--- a/app/jobs/enqueue_firm_status_checks_job.rb
+++ b/app/jobs/enqueue_firm_status_checks_job.rb
@@ -1,0 +1,53 @@
+class EnqueueFirmStatusChecksJob < ActiveJob::Base
+  include Sidekiq::Worker
+
+  # This job schedules the work carried out by FirmStatusCheckJob, which
+  # queries the status of Firms against the FCA Register API.
+  #
+  # The API has an official rate limit of 10 requests every 10 seconds, hence
+  # the values set in the constants below. We throttle requests to the API by
+  # scheduling each batch of 10 status checks to occur at the beginning of a
+  # 10-second window. The 10 checks occur concurrently and windows are
+  # sequential.
+  #
+  # The other value relevant to this functionality is the scheduled job poll
+  # interval set in the sidekiq server configuration. This controls how often
+  # Sidekiq checks for scheduled jobs. That value is ~15 seconds by defauilt -
+  # this job has been tested at that level of polling and holds up well across
+  # large numbers of firms without triggering the rate limit. Changes to that
+  # value should be tested, but in theory any value between 10 and 20 seconds
+  # should be ok. This is because in practice the API rate limit appears to be
+  # higher than the official numbers. Up to 20 requests per 10 seconds have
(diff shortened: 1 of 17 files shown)


app/jobs/enqueue_firm_status_checks_job.rb
```diff
@@ -0,0 +1,61 @@
...
+  # set in the sidekiq server configuration. This controls how often Sidekiq
+  # checks for scheduled jobs. That value should be set to 15 seconds - this
+  # job has been tested at that level of polling and holds up well across large
+  # numbers of firms without triggering the rate limit. Changes to that value
+  # should be tested, but in theory any value between 10 and 20 seconds should
+  # be ok. This is because in practice the API rate limit appears to be higher
+  # than the official numbers. Up to 20 requests per 10 seconds have been tested
+  # without issue. With a schedule poll of 15 seconds, 10 to 20 jobs should
+  # execute each time (as certain times will pick up two windows worth of jobs
+  # rather than just one, eg - the 30-second mark).
```
tomas-stefano: 👍

app/jobs/firm_status_check_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    status = response.data.fetch('Status')
+
+    if status.in? ACTIVE_FIRM_STATUS_CODES
+      logger.info "Firm #{fca_number} is ACTIVE"
+    else
+      logger.info "Firm #{fca_number} is INACTIVE with status #{status}"
+      firm = Firm.where(fca_number: fca_number).first!
+      InactiveFirm.create(firm_id: firm.id, api_status: status)
+    end
+  end
```
tomas-stefano: 👍

PR #477, opened 2019-10-23
10737 Upgrade to Rails 5.2

malcolmbaig:
[TP](https://maps.tpondemand.com/entity/10826-update-rails-and-sidekiq-versions-in)

Upgrade from 4.2 to 5.2. 

Broadly, changes are as follows:

- Removing unnecessary dependencies to ease the upgrade process.
- Stepping through versions 5.0, 5.1 and 5.2, using `rails app:upgrade` with each step to ensure config files are kept up to date.
- Fixing specs and deprecation notices.

Note that this upgrade is dependent on a Rails 5-specific branch for dough-ruby. See the commit references in the Gemfile for more detail about the required changes. There's an outstanding decision required on what we do about dough and frontend assets long-term. A spike is due to take place soon to start answering this.

See individual commits for more detail.


malcolmbaig:
Approving, as only backend developer currently online.

PR #482, opened 2019-11-25
Add travel insurance firms tab to admin

tomas-stefano:
[TP card](https://maps.tpondemand.com/entity/10765-create-firms-tab)

Some caveats:

* The travel insurance firm uses template inheritance from its
parent controller
* I added the show route but the view will be only added in a future card
(add firm profile on admin).
* The travel insurance tab won't contain some filters specific to
retirement firms


config/routes.rb
```diff
@@ -49,7 +49,9 @@
 
     resources :advisers, only: [:index, :show, :edit, :update, :destroy]
 
-    resources :firms, only: [:index, :show] do
+    resources :travel_insurance_firms, only: [:index, :show]
+
+    resources :firms, controller: 'retirement_firms', only: [:index, :show] do
```
tomas-stefano: @malcolmbaig that's the only gotcha in this PR. I decided to leave `resources :firms` as it is because I almost endup rewriting a lot of views and renaming them and there was still more and more to be renamed and changed which I kinda feel that it not part of the card of "Firms tab". Please do let me know your point of view 🙏

PR #489, opened 2019-12-12
Principals on Admin

tomas-stefano:
[TP card](https://maps.tpondemand.com/entity/10766-create-principles-tab)
[TP card](https://maps.tpondemand.com/entity/10769-create-principle-profile-page-in-admin)

There are two cards addressed in this PR since both are relevant and almost the same feature/story.

Again I decided to use template inheritance to solve the common and different logic in the views.


tomas-stefano:
Just force push to fix a rubocop offence. ❤️

PR #515, opened 2020-05-01
[TP-11382] - Page 4 create page and form

ryanharkins:
Screenshot 2020-04-30 at 16 37 59

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -8,4 +8,5 @@ require(['jquery', 'componentLoader', 'eventsWithPromises'], function ($, compon
 require(['jquery'], function ($) {
   require(['AdviserAjaxCall']);
   require(['RemoteAndFaceToFaceOptions']);
+  require(['WizardForm']);
 });
--- a/app/assets/javascripts/modules/WizardForm.js
+++ b/app/assets/javascripts/modules/WizardForm.js
@@ -1,18 +1,14 @@
 define(['jquery'], function($) {
   'use strict';
 
-  var $form = $('[data-wizard-form] form'),
-      $formItems = $form.find('.wizard-form-item'),
-      $submitButton = $form.find('button.button--primary'),
-      $prevButton = $form.find('button.button--secondary'),
+  var $form = $('[data-wizard-form] form.wizard-form'),
+      $formItems = $form.find('[data-wizard-item]'),
+      $submitButton = $form.find('[data-wizard-next-button]'),
+      $prevButton = $form.find('[data-wizard-previous-button]'),
       $currentStep = $formItems.first(),
       $currentStepNumber = 0;
 
-  $formItems.hide().first().show().addClass('show');
-  $form.addClass('js-wizard');
-
   addStepEventHandler($currentStep)
-  disableOrEnableSubmitButton($currentStep);
 
   $submitButton.on('click', function(e){
     e.preventDefault();
@@ -20,11 +16,11 @@ define(['jquery'], function($) {
     if($formItems.length - 1 == $currentStepNumber){
       $form.submit();
     }else{
-      $currentStep.hide().removeClass('show')
+      $currentStep.addClass('hidden').removeClass('show show-previous')
       $currentStepNumber += 1
 
       $currentStep = $($formItems[$currentStepNumber])
-      $currentStep.show().addClass('show')
+      $currentStep.removeClass('hidden show-previous').addClass('show')
       addStepEventHandler($currentStep)
       disableOrEnableSubmitButton($currentStep);
     }
@@ -33,16 +29,15 @@ define(['jquery'], function($) {
   $prevButton.on('click', function(e){
     e.preventDefault();
 
-    $currentStep.hide().removeClass('show-previous show')
+    $currentStep.addClass('hidden').removeClass('show-previous show')
     $currentStepNumber -= 1
 
     $currentStep = $($formItems[$currentStepNumber])
-    $currentStep.addClass('show-previous').show().addClass('show')
+    $currentStep.removeClass('hidden').addClass('show-previous')
 
     addStepEventHandler($currentStep)
     disableOrEnableSubmitButton($currentStep);
   })
-
 
   function addStepEventHandler(step){
     step.on('change', function(e){
--- a/app/assets/javascripts/modules/WizardForm.js
+++ b/app/assets/javascripts/modules/WizardForm.js
@@ -1,0 +1,39 @@
+define(['jquery'], function($) {
+  'use strict';
+
+  var $form = $('[data-wizard-form] form'),
+      $formItems = $form.find('.wizard-form-item'),
+      $submitButton = $form.find('button'),
+      $currentStep = $formItems.first(),
+      $currentStepNumber = 0;
+
+  $formItems.hide().first().show();
+  addStepEventHandler($currentStep)
+  disableOrEnableSubmitButton($currentStep);
+
+  $submitButton.on('click', function(e){
+    e.preventDefault();
+
+    if($formItems.length - 1 == $currentStepNumber){
+      $form.submit();
+    }else{
+      $currentStep.slideToggle()
+      $currentStepNumber += 1
+
+      $currentStep = $($formItems[$currentStepNumber])
+      $currentStep.slideToggle()
+      addStepEventHandler($currentStep)
+      disableOrEnableSubmitButton($currentStep);
+    }
+  })
+
+  function addStepEventHandler(step){
+    step.on('change', function(e){
+      $submitButton.attr('disabled', false)
+    })
+  }
+
+  function disableOrEnableSubmitButton(step){
+    $submitButton.attr('disabled', step.find('input:checked').length <= 0)
+  }
+});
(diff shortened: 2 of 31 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11382

masbuild:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ryanharkins:
@kiggundu part of the card was to make it work without js, this page renders all the steps underneath each other if js is disabled.

app/assets/javascripts/application.js
```diff
@@ -8,4 +8,5 @@ require(['jquery', 'componentLoader', 'eventsWithPromises'], function ($, compon
 require(['jquery'], function ($) {
   require(['AdviserAjaxCall']);
   require(['RemoteAndFaceToFaceOptions']);
+  require(['WizardForm']);
```
davidtrussler, line 11: Generally on MAS projects we load JS components via our Dough Component Loader mechanism rather than through the manifest in this way. That would be our preferred way of doing this but since in this project there appears to already be a mixture of both methods it's probably not critical.

app/views/travel_insurance_registrations/medical_conditions_questionaire_form.html.erb
```diff
@@ -0,0 +1,73 @@
+<%= heading_tag(t(registration_title), level: 1) %>
+<p><%= t('travel_insurance_registrations.intro') %></p>
+
+<h3>Step 4 of 4 </h3>
+
+<div class="l-2col-even-row">
+  <div class="l-2col">
+    <section class="l-2col-main" data-wizard-form>
+      <%= form_for @form, url: medical_conditions_questionaire_travel_insurance_registrations_path, html: { class: 'form' }, builder: Dough::Forms::Builders::Validation do |f| %>
```
davidtrussler, line 9: We'd prefer to use a more descriptive classname here than `form`. If we are applying styles to forms in general we can use the element name but we prefer to be more specific than that.

app/views/travel_insurance_registrations/medical_conditions_questionaire_form.html.erb
```diff
@@ -0,0 +1,73 @@
+<%= heading_tag(t(registration_title), level: 1) %>
+<p><%= t('travel_insurance_registrations.intro') %></p>
+
+<h3>Step 4 of 4 </h3>
```
davidtrussler, line 4: I don't think, semantically-speaking, this is a heading. I'd prefer to see it marked up as a `<p>` with a classname that reflects its role on the page. If it is a heading it should be a `<h2>` as to skip the numerical ordering of headings can be confusing to assistive technology.

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
+.form.js-wizard{
```
davidtrussler, line 1: We prefer not to concatenate classnames but rather to use nesting in this instance. However the use of the `js-wizard` class is not really necessary here since we are using Modernizr to attach a `js` class to the `html` element and our preferred method of defining classes for JS components is to wrap the styles in a `.js` container in the SASS.

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
...
+  position: relative;
+  overflow: hidden;
```
davidtrussler, line 3: I don't think this is strictly necessary since the items are not shown when not in an "active" state. It also means there is no need for the padding value below to avoid the cut-off of the radio element.

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
...
+
+  .wizard-form-item{
+    transform: translateX(100%);
+    padding: 1px;
```
davidtrussler, line 7: See above on this value.

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
...
+    background-color: #fff;
```
davidtrussler, line 8: We prefer to use a variable `$color-white` here in place of the value `#fff`. 
But in addition is the colour specification necessary here?

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  .wizard-form-item.show-previous {
```
davidtrussler, line 11: As above, we prefer to nest these classes over concatenating them, e.g. 
```
.wizard-form-item {
  &.show-previous {}
}
```

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
```diff
@@ -0,0 +1,20 @@
...
+    transform: translateX(-100%);
+  }
+
+  .wizard-form-item.show {
```
davidtrussler, line 15: As above

app/views/travel_insurance_registrations/medical_conditions_questionaire_form.html.erb
```diff
@@ -0,0 +1,73 @@
...
+                <label>
+                  <%= f.radio_button key, 0, class: 'form__group-input' %>
+                  <%= t('registration.answer_no') %>
+                </label>
+              </div>
+            </fieldset>
+          <% end %>
+        <% end %>
+
+        <button class="button button--secondary" style="display:none"><%= t('travel_insurance_registrations.previous_question_button') %></button>
```
davidtrussler, line 54: I think we should avoid using `button--secondary` as a class for this since that name has a more specific semantic role. Instead we should use a bespoke class name for this.

davidtrussler, line 54: Can we avoid inline styles here? Our preference is to control this from stylesheets.

app/assets/javascripts/modules/WizardForm.js
```diff
@@ -0,0 +1,61 @@
...
+  'use strict';
+
+  var $form = $('[data-wizard-form] form'),
+      $formItems = $form.find('.wizard-form-item'),
+      $submitButton = $form.find('button.button--primary'),
+      $prevButton = $form.find('button.button--secondary'),
+      $currentStep = $formItems.first(),
+      $currentStepNumber = 0;
+
+  $formItems.hide().first().show().addClass('show');
```
davidtrussler, line 11: Our preference is to keep the hidden/shown states of these items in the CSS by adding classes to the element rather than directly setting their display in the JS.

app/assets/javascripts/modules/WizardForm.js
```diff
@@ -0,0 +1,61 @@
...
+  $form.addClass('js-wizard');
```
davidtrussler, line 12: We don't need to add this class here if we use the `js` hook in the `html` element.

app/assets/javascripts/modules/WizardForm.js
```diff
@@ -0,0 +1,61 @@
+define(['jquery'], function($) {
+  'use strict';
+
+  var $form = $('[data-wizard-form] form'),
+      $formItems = $form.find('.wizard-form-item'),
+      $submitButton = $form.find('button.button--primary'),
+      $prevButton = $form.find('button.button--secondary'),
```
davidtrussler, line 7: we prefer to use `data-` values to find elements in the DOM instead of using classes, which should be reserved for styles.

davidtrussler:
I've left a few comments regarding certain aspects of preferred practice. 

In addition to those 
- we would normally unit test our JS components using Karma as our test framework (which is already implemented here on other components)
- we need to consider some accessibility concerns. We could either 
  - consider all the questions to be visible to assistive technology, or
  - add aria attributes to describe the hidden/shown state of the question. 
- other accessibility issues we should consider here would include the moving the focus so that users navigating by keyboard are able to move successfully through the questions. 

These concerns could be added to a further card if that would make more sense.

app/assets/javascripts/modules/WizardForm.js
ryanharkins: Agreed, I moved initial showing/hiding in css now

app/views/travel_insurance_registrations/medical_conditions_questionaire_form.html.erb
ryanharkins: Ok, I gave it a new class name and extended it with `button--secondary` styling

app/assets/stylesheets/components/common/_travel_insurance_wizard_form.scss
ryanharkins: Removed, it was not required anymore. I was using it when sliding in the element so there wasn't text on text

app/assets/javascripts/modules/WizardForm.js
ryanharkins: Removed

app/views/travel_insurance_registrations/medical_conditions_questionaire_form.html.erb
ryanharkins: Updated to use a `p` tag

PR #516, opened 2020-05-01
[TP-10993] - Signup form page 4 logic

ryanharkins:


jordanlg27:
https://maps.tpondemand.com/entity/10993

masbuild:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/controllers/travel_insurance_registrations_controller.rb
```diff
@@ -101,6 +104,11 @@ def medical_conditions_questionaire_form_params
     )
   end
 
+  def medical_questionaire_acceptable?
+    positive_answers_count = medical_conditions_questionaire_form_params.values.select { |i| i == '1' }.count
```
kiggundu, line 108: I think we should use `true|false` as opposed to `1|0` for boolean answers. Makes ERB and form code easier to read and maps nicely with boolean flag persisted DB fields. I noticed that's not the trend in the existing RAD code though :-/

kiggundu: Considering the feedback that this will be an involving change especially wrt test maintenance and also that the existing code is written this way, I believe the implied improved code aesthetics are not important enough to hold this card back.

PR #561, opened 2020-09-21
Make sure we re-index travel firms when their office gets updated

ryanharkins:


masbuild:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/models/office.rb
```diff
@@ -53,6 +53,8 @@ class Office < ApplicationRecord
   def notify_indexer
     if officeable_type == 'Firm'
       UpdateAlgoliaIndexJob.perform_later(model_name.name, id, officeable_id)
+    elsif officeable
+      officeable.notify_indexer
     end
```
kiggundu, line 58: I don't understand this. Id have thought its alot clearer to replace `officeablel.notify_indexer` with a literal call to `UpdateAlgoliaIndexJob...` as well without the extra parameter. Or alternatively move both calls to within `officeable.notify_indexer`. Is there something about the way it works that i've misunderstood?

kiggundu (commented):
Left a comment wrt clarifying the code around indexing firms

app/models/office.rb
ryanharkins: Offices in RAD have their own index in Algolia, so they have an `AlgoliaIndex::Office` model, Travel does not use a separate index because FE team requests the index to be in a particular format. As there is no index for `offices` we reindex the `TravelInsuranceFirm` model which includes the nested office information. 

Looking at it now, maybe we could move the logic in `AlgoliaIndex::Office` instead

thenapking:
I have reviewed this, and work on another ticket will supercede it.

app/models/office.rb
thenapking: This logic has been added in because the structure of the application has become too complicated.  Travel Firm Offerings should move to a separate model, which feeds the index.  We should look to see how we can standardise the firms listed by using either the Office or Firm model. The officable relation, and firmable are unnecessarily complex.  We should avoid polymorphic relations.  The Firm model should be refactored to use STI, which would remove the need for complex relations.

PR #568, opened 2020-12-16
TP 11913 firm status check error

thenapking:
On staging and production some tables are missing auto-incrementing sequences which is causing errors.  The schema is correct.  It's possible this is pg issue. 
Create missing sequences and alter tables, but not alter the schema.


jordanlg27:
https://maps.tpondemand.com/entity/11913

thenapking:
This second pull request is not required.  It will alter schema rather than fix issue

PR #579, opened 2021-02-15
TP-11965 Undefined Method in Firm Status Check

thenapking:
[TP-11965](https://maps.tpondemand.com/entity/11965-rad-undefined-method-in-firm-status)

Catch error when there is no response from API on checking a firm.


jordanlg27:
https://maps.tpondemand.com/entity/11965

PR #580, opened 2021-02-17
TP-11815 Add 1 - 3 Months Duration

thenapking:
[TP-11815](https://maps.tpondemand.com/entity/11815-tad-admin-add-1-3-months)

Clean up rubocop items left over from TP-11770
Add :land_50_days_max_age,   :cruise_50_days_max_age columns for 3-6 months duration
Update logic in TripCover model so that single trips only are complete when new columns are not null
Update view
Update tests
Add cols to Algolia index


jordanlg27:
https://maps.tpondemand.com/entity/11815

